PR #117, opened 2023-05-24
chore: fix msvc compiler warnings related to strcasecmp

cwaldren-ld:
If we don't define the symbol, we don't need to undefine it.

### DIFF
--- a/libs/internal/src/network/http_requester.cpp
+++ b/libs/internal/src/network/http_requester.cpp
@@ -1,11 +1,23 @@
 #include <boost/url.hpp>
 
+#include <algorithm>
+#include <cctype>
 #include <optional>
 #include <utility>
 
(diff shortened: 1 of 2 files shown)


libs/internal/src/network/http_requester.cpp
```diff
@@ -1,12 +1,24 @@
 #include <boost/url.hpp>
 
+#include <algorithm>
+#include <cctype>
```
kinyoklion, line 4: Nothing is actually from this header?

kinyoklion (approved):

libs/internal/src/network/http_requester.cpp
cwaldren-ld: Woops, remains of trying to fix it with `std::transform`.

PR #119, opened 2023-05-26
feat: Add c bindings for FlagNotifier.

kinyoklion:


libs/client-sdk/tests/client_c_bindings_test.cpp
```diff
@@ -20,3 +20,41 @@ TEST(ClientBindings, MinimalInstantiation) {
...
+void ListenerFunction(char const* flag_key,
+                      LDValue new_value,
+                      LDValue old_value,
+                      bool deleted,
+                      void* user_data) {
+}
+
+// This test registers a listener. It doesn't use the listener, but it
+// will at least ensure 1.) Compilation, and 2.) Allow sanitizers to run.
+TEST(ClientBindings, RegisterFlagListener) {
```
kinyoklion, line 33: When we do TestData for the server possibly we can add something similar to the client. It would make legitimate testing of these much simpler. 

On the C++ side it is better tested because we can directly make the updater.

kinyoklion: I did test this a number of ways in the hello-c app.

libs/common/include/launchdarkly/connection.hpp
```diff
@@ -1,6 +1,6 @@
 #pragma once
 
-namespace launchdarkly::client_side {
+namespace launchdarkly {
```
kinyoklion, line 3: This was moved to common, but client_side was not removed.

libs/client-sdk/src/bindings/c/sdk.cpp
```diff
@@ -291,5 +291,37 @@ LD_EXPORT(void) LDClientSDK_Free(LDClientSDK sdk) {
...
 }
 
+LD_EXPORT(LDListenerConnection)
+LDClientSDK_FlagNotifier_OnFlagChange(LDClientSDK sdk,
+                                      char const* flag_key,
+                                      struct LDFlagListener listener) {
+    LD_ASSERT_NOT_NULL(sdk);
+    LD_ASSERT_NOT_NULL(flag_key);
+
+    auto connection = TO_SDK(sdk)->FlagNotifier().OnFlagChange(
```
cwaldren-ld, line 301: If `listener.FlagChanged` is null, then is there any need to register a callback?

kinyoklion: I guess we could re-word the documentation to indicate that the returned connection can be NULL if no listener is registered.

Currently the need to register a listener is just to get the connection.

cwaldren-ld (approved):

PR #120, opened 2023-05-30
fix: update required Boost version to 1.81 for Boost.URL

cwaldren-ld:
We use Boost.URL, which was [introduced](https://www.boost.org/users/history/version_1_81_0.html) in 1.81.

Potentially we could switch to foxy's built-in URL library if we wanted to relax the boost requirement, although 1.81 came out in December 2022 so it's not exactly bleeding edge.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,29 +3,30 @@
 This repository contains LaunchDarkly SDK packages which are written in C++.
 This includes shared libraries, used by SDKs and other tools, as well as SDKs.
 
-This repository contains beta software and should not be considered ready for production use while this message is visible.
+This repository contains beta software and should not be considered ready for production use while this message is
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -34,7 +35,7 @@ This repository contains beta software and should not be considered ready for pr
 1. C++17 and above
 1. CMake 3.19 or higher
 1. Ninja (if using the included build scripts)
-1. Boost version 1.80 or higher.
+1. Boost version 1.81 or higher
 1. OpenSSL
```
cwaldren-ld, line 39: I only made this change, the rest was auto-format by the IDE.

kinyoklion (approved):

PR #122, opened 2023-05-30
feat: add Version method to obtain SDK version

cwaldren-ld:
Adds a `Version/LDClientSDK_Version`method, which exposes the client's version via function call.

In the previous way, it was a public static class member. This way, we have a bit more encapsulation and can also expose it via C binding.

### DIFF
--- a/libs/client-sdk/include/launchdarkly/client_side/client.hpp
+++ b/libs/client-sdk/include/launchdarkly/client_side/client.hpp
@@ -242,12 +242,6 @@ class IClient {
      */
     virtual flag_manager::IFlagNotifier& FlagNotifier() = 0;
 
-    /**
-     * Returns the version of the SDK.
-     * @return String representing version of the SDK.
-     */
-    [[nodiscard]] virtual char const* Version() const = 0;
-
     virtual ~IClient() = default;
     IClient(IClient const& item) = delete;
     IClient(IClient&& item) = delete;
@@ -318,7 +312,11 @@ class Client : public IClient {
 
     flag_manager::IFlagNotifier& FlagNotifier() override;
 
-    [[nodiscard]] char const* Version() const override;
+    /**
+     * Returns the version of the SDK.
+     * @return String representing version of the SDK.
+     */
+    [[nodiscard]] static char const* Version();
 
    private:
     inline static char const* const kVersion =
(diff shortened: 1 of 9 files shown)


libs/client-sdk/tests/client_c_bindings_test.cpp
```diff
@@ -18,5 +18,9 @@ TEST(ClientBindings, MinimalInstantiation) {
 
     LDClientSDK sdk = LDClientSDK_New(config, context);
 
+    char const* version = LDClientSDK_Version(sdk);
+    ASSERT_TRUE(version);
+    ASSERT_STRNE(version, "");
```
kinyoklion, line 23: You could have release-please update these if you want.

libs/client-sdk/include/launchdarkly/client_side/client.hpp
```diff
@@ -242,6 +242,12 @@ class IClient {
      */
     virtual flag_manager::IFlagNotifier& FlagNotifier() = 0;
 
+    /**
+     * Returns the version of the SDK.
+     * @return String representing version of the SDK.
+     */
+    virtual char const* Version() const = 0;
```
kinyoklion, line 249: `[[nodiscard]]`

kinyoklion (approved):
Approved with minor comments.

kinyoklion (approved):
I do think a public static method would be more convenient. Because then you don't need to create a client to know the SDK version.

cwaldren-ld:
Agreed. Will updated.

libs/client-sdk/tests/client_c_bindings_test.cpp
```diff
@@ -18,5 +18,9 @@ TEST(ClientBindings, MinimalInstantiation) {
 
     LDClientSDK sdk = LDClientSDK_New(config, context);
 
+    char const* version = LDClientSDK_Version();
+    ASSERT_TRUE(version);
+    ASSERT_STREQ(version, "0.1.0");  // {x-release-please-version}
```
kinyoklion, line 23: Need to add these files to release-please-config.json as well.

It only looks at specific files to try to update them.
```
      "extra-files": [
        "include/launchdarkly/client_side/client.hpp"
      ]
```

cwaldren-ld: Ah, missed that.

kinyoklion (approved):

kinyoklion (approved):

PR #130, opened 2023-05-31
feat: Allow for easier creation of contexts from existing contexts.

kinyoklion:


libs/common/tests/context_builder_test.cpp
```diff
@@ -176,4 +175,70 @@ TEST(ContextBuilderTests, AccessKindBuilderMultipleTimes) {
...
+TEST(ContextBuilderTests, AddAttributeToExistingContext) {
+    auto context = ContextBuilder()
+                       .Kind("user", "potato")
+                       .Name("Bob")
+                       .Set("city", "Reno")
+                       .SetPrivate("private", "a")
+                       .Build();
+
+    auto builder = ContextBuilder(context);
+    if (auto updater = builder.Kind("user")) {
```
kinyoklion, line 187: @mmrj Adding "state" and "sneaky" to the context.

libs/common/tests/context_builder_test.cpp
```diff
@@ -176,4 +175,70 @@ TEST(ContextBuilderTests, AccessKindBuilderMultipleTimes) {
...
+}
+
+TEST(ContextBuilderTests, AddKindToExistingContext) {
+    auto context = ContextBuilder()
+                       .Kind("user", "potato")
+                       .Name("Bob")
+                       .Set("city", "Reno")
+                       .Build();
+
+    auto updated_context =
```
kinyoklion, line 210: @mmrj Adding an "org" to the context.

libs/common/include/launchdarkly/attributes_builder.hpp
```diff
@@ -155,17 +180,39 @@ class AttributesBuilder final {
         return *this;
     }
 
+    /**
+     * Start adding a kind to the context.
```
mmrj, line 184: Wording nit: I think you are adding a new context with the specified kind to the builder ?

libs/common/include/launchdarkly/attributes_builder.hpp
```diff
@@ -155,17 +180,39 @@ class AttributesBuilder final {
...
     AttributesBuilder& Kind(std::string kind, std::string key) {
         return builder_.Kind(kind, key);
     }
 
     /**
-     * Build the context. This method should not be called more than once.
-     * It moves the builder content into the built context.
+     * Start updating an existing kind.
+     *
+     * @param kind The kind to start updating.
```
mmrj, line 202: same as above, I think you're updating the existing context with the given kind

libs/common/include/launchdarkly/context_builder.hpp
```diff
@@ -79,24 +92,27 @@ class ContextBuilder final {
                                                      std::string key);
 
     /**
-     * Build a context. This should only be called once, because
-     * the contents of the builder are moved into the created context.
+     * Start updating an existing kind.
      *
-     * After the context is build, then this builder, and any kind builders
-     * associated with it, should no longer be used.
+     * @param kind The kind to start updating.
```
mmrj, line 97: same here

I think the reason I find the current wording a little confusing is because in the UI a context kind is an entity that you can create and define (Contexts list --> Kind tab --> Create kind). It's separate from any context where it's used, and this is definitely not working with that.

libs/common/tests/context_builder_test.cpp
mmrj: nice!

mmrj: could I also do:
```
updater->Set("city", "Winnemucca")
```
and overwrite the `Reno` value?

libs/common/tests/context_builder_test.cpp
mmrj: nice. Do I need to call `IdentifyAsync` with `updated_context` after this?

mmrj (commented):
Semantics of using this LGTM. I left a couple of questions and some nits on the comments, but nothing blocking.

libs/common/tests/context_builder_test.cpp
kinyoklion: Yes, this just makes a context. You need to identify with that context to get flag values for it.

libs/common/tests/context_builder_test.cpp
kinyoklion: Yes.

libs/common/include/launchdarkly/attributes_builder.hpp
kinyoklion: (The kind happens to be a context, like the address happens to be a string, or the name is a string.)

libs/common/include/launchdarkly/context_builder.hpp
kinyoklion: Yeah, adding a "org" to a context is different than adding an "org" to your launchdarky account. The "org" you put into a context has attributes, and you cannot even make it without any attributes.

kinyoklion: Both are kinds though.

mmrj: Yeah I think got hung up on the two uses of "kind":
 * the entity you add to your LD account, can (but don't have to) use when defining a new context in the SDK, and can get a list of using the REST API, etc.
 * the thing that, as you say above, happens to be a context, like the address happens to be a string, or the name is a string

PR #145, opened 2023-06-08
fix: ensure x-launchdarkly-tags is sent in event requests

cwaldren-ld:
Moves application tag into HTTP Config, so it can propagate to data sources + event sender.

### DIFF
--- a/libs/common/src/config/http_properties_builder.cpp
+++ b/libs/common/src/config/http_properties_builder.cpp
@@ -72,8 +72,12 @@ HttpPropertiesBuilder<SDK>& HttpPropertiesBuilder<SDK>::Headers(
 template <typename SDK>
 HttpPropertiesBuilder<SDK>& HttpPropertiesBuilder<SDK>::Header(
     std::string key,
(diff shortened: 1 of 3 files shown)


libs/common/src/config/http_properties_builder.cpp
```diff
@@ -72,8 +72,12 @@ HttpPropertiesBuilder<SDK>& HttpPropertiesBuilder<SDK>::Headers(
...
 HttpPropertiesBuilder<SDK>& HttpPropertiesBuilder<SDK>::Header(
     std::string key,
-    std::string value) {
-    base_headers_.insert_or_assign(key, value);
+    std::optional<std::string> value) {
+    if (value) {
+        base_headers_.insert_or_assign(key, *value);
+    } else {
+        base_headers_.erase(key);
+    }
```
cwaldren-ld, line 80: Could also have this just be a no-op if it's nullopt, but this also allows for removal without needing another API.

kinyoklion (approved):

PR #150, opened 2023-06-13
feat: update event processor to handle context key deduplication

cwaldren-ld:
Augments event processor with context-key deduplication abilities using an LRU cache.

Since the implementation is shared between server/client, I use `if constexpr` on the SDK template to enable the behavior only on server-side SDKs.

Cache eviction behavior subject to ongoing discussion.


libs/internal/src/events/asio_event_processor.cpp
```diff
@@ -212,47 +213,63 @@ std::vector<OutputEvent> AsioEventProcessor<SDK>::Process(
...
+
+                if constexpr (std::is_same<SDK,
+                                           config::shared::ServerSDK>::value) {
+                    if (!context_key_cache_.Notice(
+                            event.context.CanonicalKey())) {
+                        out.emplace_back(client::IdentifyEvent{
+                            event.creation_date,
+                            filter_.filter(event.context)});
+                    }
+                }
```
cwaldren-ld, line 228: This is the actual change (1/2), not sure why the diff chunk is so large.

libs/internal/src/events/asio_event_processor.cpp
```diff
@@ -212,47 +213,63 @@ std::vector<OutputEvent> AsioEventProcessor<SDK>::Process(
...
-                   }},
+        overloaded{
+            [&](client::FeatureEventParams&& event) {
+                summarizer_.Update(event);
+
+                if constexpr (std::is_same<SDK,
+                                           config::shared::ServerSDK>::value) {
+                    if (!context_key_cache_.Notice(
+                            event.context.CanonicalKey())) {
+                        out.emplace_back(client::IdentifyEvent{
```
kinyoklion, line 224: This should be an index event?

libs/internal/include/launchdarkly/events/context_key_cache.hpp
```diff
@@ -0,0 +1,41 @@
+#pragma once
+#include <list>
+#include <string>
+#include <unordered_map>
+namespace launchdarkly::events {
+
+class ContextKeyCache {
```
kinyoklion, line 7: Should this be a little more generic? The cache big segments needs is mostly the same, except it is by user hash (and actually needs the expiration).

cwaldren-ld: Ah, yea might as well parameterize it now. I'll leave out the expiration until we decide.

libs/internal/src/events/asio_event_processor.cpp
cwaldren-ld: Yep, fixed.

libs/internal/src/events/context_key_cache.cpp
```diff
@@ -0,0 +1,32 @@
+#include <launchdarkly/events/lru_cache.hpp>
```
kinyoklion, line 1: Filename of this one is still context_key_cache. 
But maybe if this cache is only for strings, then it doesn't need to be generic.

(Big segments, if we do it at some point has three states. Basically a tribool. Included in the segment, not included in the segment, or it doesn't exist. So when it isn't part of the segment we still cache that.)

cwaldren-ld: Uhg, sorry for the churn. Renamed to `lru_cache` and renamed the test file as well.

kinyoklion: No biggie.

kinyoklion (approved):

PR #152, opened 2023-06-14
feat: deserialize segments

cwaldren-ld:
This PR extends our existing deserialization/serialization technique from the client-side SDK to the server-side SDK data model. 

Specifically, it adds deserialization for the segment structure. I've also introduced some helpers to make the logic less repetitive, and centralize some decisions (like what to do if a JSON field isn't present, or is null.)

As background, the `tag_invoke` functions present in this PR are what allows the `boost::json` library to recursively deserialize complex structures. 

Here, I've added a couple new `tag_invokes` for bool, string, vector, etc. The special part is that they deserialize into `expected<type, error>` rather than the primitive directly; this allows us to return an error if the deserialization failed.

### DIFF
--- a/libs/internal/include/launchdarkly/data_model/item_descriptor.hpp
+++ b/libs/internal/include/launchdarkly/data_model/item_descriptor.hpp
@@ -61,48 +61,4 @@ template <typename T>
 ItemDescriptor<T>::ItemDescriptor(T item)
     : version(item.Version()), item(std::move(item)) {}
 
-template <typename T>
-tl::expected<std::unordered_map<std::string, ItemDescriptor<T>>, JsonError>
-tag_invoke(boost::json::value_to_tag<
-               tl::expected<std::unordered_map<std::string, ItemDescriptor<T>>,
-                            JsonError>> const& unused,
-           boost::json::value const& json_value) {
-    boost::ignore_unused(unused);
-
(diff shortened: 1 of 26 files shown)


libs/internal/include/launchdarkly/data_model/item_descriptor.hpp
```diff
@@ -61,48 +61,4 @@ template <typename T>
 ItemDescriptor<T>::ItemDescriptor(T item)
     : version(item.Version()), item(std::move(item)) {}
```
cwaldren-ld, line 63: This change is just moving these `tag_invokes` into the `serialization` folder.

libs/internal/include/launchdarkly/data_model/sdk_data_set.hpp
```diff
@@ -0,0 +1,20 @@
...
+#include <optional>
+#include <tl/expected.hpp>
+#include <unordered_map>
+
+namespace launchdarkly::data_model {
+
+struct SDKDataSet {
+    using FlagKey = std::string;
+    using SegmentKey = std::string;
+    // std::unordered_map<FlagKey, ItemDescriptor<Flag>> flags;
```
cwaldren-ld, line 15: Will be added if this deserialization logic is amenable.

PR #153, opened 2023-06-22
feat: segment data model

cwaldren-ld:
This PR extends our existing deserialization/serialization technique from the client-side SDK to the server-side SDK data model. 

Specifically, it adds deserialization for the segment structure. I've also introduced some helpers to make the logic less repetitive, and centralize some decisions (like what to do if a JSON field isn't present, or is null.)

As background, the `tag_invoke` functions present in this PR are what allows the `boost::json` library to recursively deserialize complex structures. 

Here, I've added a couple new `tag_invokes` for bool, string, vector, etc. The special part is that they deserialize into `expected<type, error>` rather than the primitive directly; this allows us to return an error if the deserialization failed.

### DIFF
--- a/libs/client-sdk/src/flag_manager/flag_persistence.cpp
+++ b/libs/client-sdk/src/flag_manager/flag_persistence.cpp
@@ -4,6 +4,7 @@
 #include <launchdarkly/encoding/sha_256.hpp>
 
 #include <launchdarkly/serialization/json_evaluation_result.hpp>
+#include <launchdarkly/serialization/json_item_descriptor.hpp>
 
 #include <utility>
 
--- a/libs/client-sdk/src/flag_manager/flag_persistence.cpp
+++ b/libs/client-sdk/src/flag_manager/flag_persistence.cpp
@@ -5,6 +5,7 @@
 
 #include <launchdarkly/serialization/json_evaluation_result.hpp>
 #include <launchdarkly/serialization/json_item_descriptor.hpp>
+#include <launchdarkly/serialization/json_primitives.hpp>
 
 #include <utility>
 
@@ -20,6 +21,17 @@ std::string PersistenceEncodeKey(std::string const& input) {
 static std::string MakeEnvironment(std::string const& prefix,
                                    std::string const& sdk_key) {
     return prefix + "_" + PersistenceEncodeKey(sdk_key);
+}
+
+std::unordered_map<std::string, ItemDescriptor> remove_nulls(
+    std::unordered_map<std::string, std::optional<ItemDescriptor>> map) {
+    std::unordered_map<std::string, ItemDescriptor> result;
+    for (auto [key, value] : map) {
+        if (value.has_value()) {
+            result.emplace(key, std::move(value.value()));
+        }
+    }
+    return result;
 }
 
 FlagPersistence::FlagPersistence(std::string const& sdk_key,
@@ -74,8 +86,9 @@ void FlagPersistence::LoadCached(Context const& context) {
     }
 
     auto res = boost::json::value_to<tl::expected<
-        std::unordered_map<std::string,
-                           launchdarkly::client_side::ItemDescriptor>,
+        std::optional<std::unordered_map<
+            std::string,
+            std::optional<launchdarkly::client_side::ItemDescriptor>>>,
         JsonError>>(parsed);
     if (!res) {
         LD_LOG(logger_, LogLevel::kError)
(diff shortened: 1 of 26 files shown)


libs/client-sdk/src/flag_manager/flag_persistence.cpp
```diff
@@ -21,6 +23,17 @@ static std::string MakeEnvironment(std::string const& prefix,
     return prefix + "_" + PersistenceEncodeKey(sdk_key);
 }
 
+std::unordered_map<std::string, ItemDescriptor> remove_nulls(
```
kinyoklion, line 26: nit: Not a big fan of this being duplicated.

libs/client-sdk/tests/data_source_event_handler_test.cpp
```diff
@@ -90,7 +90,7 @@ TEST(StreamingDataHandlerTests, BadSchemaPut) {
         ContextBuilder().Kind("user", "user-key").Build(), test_handler, logger,
         status_manager);
 
-    auto res = stream_handler.HandleMessage("put", "{\"potato\": {}}");
+    auto res = stream_handler.HandleMessage("put", "{\"potato\": 3}");
```
kinyoklion, line 93: Can I get an explainer on this one?

libs/common/include/launchdarkly/attribute_reference.hpp
```diff
@@ -15,8 +15,9 @@ namespace launchdarkly {
  * launchdarkly::Context::Get, or to identify an attribute or nested value that
  * should be considered private
  * with launchdarkly::AttributesBuilder<BuilderReturn, BuildType>::SetPrivate or
- * launchdarkly::AttributesBuilder<BuilderReturn, BuildType>::AddPrivateAttribute
- * (the SDK configuration can also have a list of private attribute references).
+ * launchdarkly::AttributesBuilder<BuilderReturn,
```
kinyoklion, line 18: I think that the auto formatter bungled this, and that it may cause problems with the docs.

libs/internal/include/launchdarkly/data_model/sdk_data_set.hpp
```diff
@@ -0,0 +1,21 @@
...
+#include <optional>
+#include <tl/expected.hpp>
+#include <unordered_map>
+
+namespace launchdarkly::data_model {
+
+struct SDKDataSet {
+    using FlagKey = std::string;
+    using SegmentKey = std::string;
+    // std::unordered_map<FlagKey, ItemDescriptor<Flag>> flags;
```
kinyoklion, line 15: Commented out code.

libs/internal/include/launchdarkly/data_model/segment.hpp
```diff
@@ -0,0 +1,69 @@
+#pragma once
+
+#include <boost/json/value.hpp>
```
kinyoklion, line 3: Could we group these into internal dependencies, boost dependencies, and std deps?

libs/internal/include/launchdarkly/data_model/segment.hpp
```diff
@@ -0,0 +1,69 @@
...
+#include <launchdarkly/value.hpp>
+#include <optional>
+#include <string>
+#include <tl/expected.hpp>
+#include <unordered_map>
+#include <vector>
+
+namespace launchdarkly::data_model {
+
+struct Segment {
```
kinyoklion, line 14: I think some blank lines between struct definitions would improve the readability.

kinyoklion (approved):

libs/client-sdk/src/flag_manager/flag_persistence.cpp
cwaldren-ld: Forgot about this. Will fix.

libs/internal/include/launchdarkly/data_model/sdk_data_set.hpp
cwaldren-ld: (This one is added in the next (flag model) PR)

kinyoklion (approved):

kinyoklion (approved):

libs/common/include/launchdarkly/attribute_reference.hpp
cwaldren-ld: Haha, it was one character too long - I removed the gratuitous space for the template params and fixed it that way.

libs/client-sdk/tests/data_source_event_handler_test.cpp
cwaldren-ld: Reverted. 

This was due to some confusion on my part - I intended to make `{}` be a valid message (viewing it as an implicit value of `0` for `version`, and `null` for `value`.) 

But I implemented that logic incorrectly and made it a valid deserialization because I was treating it as empty/omitted. 

That's a backwards way of thinking about it: "it's an empty object, therefore treat it as empty/omitted" vs "it's omitted/omitted, therefore treat it as an empty object" - the former is wrong, the latter is what I'm doing in the later Flag Model PR.

PR #156, opened 2023-06-27
feat: flag data model

cwaldren-ld:
This PR builds on #153, extending the deserialization to flag data. 

I've added to the `PARSE_FIELD` macro to handle the following cases:
1. Parse a field where the zero-value is a valid part of the domain: `PARSE_FIELD`. In other words, if the field is omitted or null, it will be filled in with a default value. Most common scenario.
2. Like (1), but supply a default value in case the field is null or omitted: `PARSE_FIELD_DEFAULT`.
3. Parse a field that doesn't have a valid default value, and so is represented as an `optional`:  `PARSE_CONDITIONAL_FIELD`. For example, the value `""` wouldn't be a valid default for a unique key. 
4. Parse a field that must be present in the JSON data: `PARSE_REQUIRED_FIELD`. Usage of this should be rare; the test would be if the SDK is totally unable to function without this. 

To achieve this, I've made the `tag_invoke`s operate on `expected<optional<T>, JsonError`'s, instead of `expected<T, JsonError`.

We could potentially revert that (and instead add a new enum case to `JsonError`), but I figure `optional` is a good representation of null/omitted. I think I need to actually try removing it to determine if it results in a better API.

One thing lacking is better error messages - we use kSchemaFailure everywhere but we now have the ability to report exactly what was missing/wrong type, etc.


libs/internal/tests/data_model_serialization_test.cpp
```diff
@@ -27,7 +29,8 @@ TEST(SDKDataSetTests, DeserializesZeroSegments) {
         boost::json::value_to<tl::expected<data_model::SDKDataSet, JsonError>>(
```
cwaldren-ld, line 29: Some of these tests change in the (not yet issued) evaluator PR, due to slight changes in the data model.

libs/common/tests/value_test.cpp
```diff
@@ -4,11 +4,11 @@
 #include <string>
 #include <vector>
 
+#include <boost/json.hpp>
```
kinyoklion, line 7: Could we put a blank line between boost deps and LD deps.

libs/internal/include/launchdarkly/data_model/flag.hpp
```diff
@@ -0,0 +1,85 @@
+#pragma once
+
+#include <boost/json/value.hpp>
```
kinyoklion, line 3: Splitting these into groups would be nice.

libs/internal/src/serialization/json_attributes.cpp
```diff
@@ -1,3 +1,5 @@
+#include <boost/core/ignore_unused.hpp>
+#include <boost/json.hpp>
 #include <launchdarkly/serialization/json_attributes.hpp>
```
kinyoklion, line 3: Can we put a blank line here.

libs/internal/src/serialization/json_context.cpp
```diff
@@ -1,6 +1,8 @@
+#include <boost/json.hpp>
 #include <launchdarkly/context_builder.hpp>
```
kinyoklion, line 2: Blank line.

libs/internal/src/serialization/json_evaluation_reason.cpp
```diff
@@ -1,6 +1,7 @@
+#include <boost/json.hpp>
+#include <launchdarkly/serialization/json_errors.hpp>
```
kinyoklion, line 2: Blank line.

libs/internal/src/serialization/json_evaluation_result.cpp
```diff
@@ -1,10 +1,11 @@
+#include <boost/core/ignore_unused.hpp>
+#include <boost/json.hpp>
+#include <launchdarkly/serialization/json_errors.hpp>
```
kinyoklion, line 3: Blank line.

kinyoklion (approved):

libs/internal/include/launchdarkly/data_model/flag.hpp
cwaldren-ld: Clang format is doing something smart. But I guess not smart enough, I haven't paid attention to how exactly it is ordering.

kinyoklion: It organizes things within any existing groups. It just alphabetizes things. In lots of the other files I had split the groups, then it will help maintain them.

kinyoklion: It isn't strictly needed, but I think it makes life nice in the long-term.

libs/internal/src/serialization/json_evaluation_result.cpp
cwaldren-ld: sorry :(

All fixed.

kinyoklion (approved):

libs/internal/src/serialization/json_evaluation_result.cpp
kinyoklion: No worries.

libs/internal/tests/data_model_serialization_test.cpp
kinyoklion: Yeah, I am depending a bit on testing for that. Mainly checked conditions matched manes.

kinyoklion (approved):

PR #157, opened 2023-06-30
fix: Fix compilation with boost 1.82.

kinyoklion:
This PR fixes compilation using boost 1.82.

It also fixes the CI issues from boost 1.81 being removed from brew, and floated with chocolatey.

Further changes, in another PR, will be needed to improve boost dependency robustness.

### DIFF
--- a/.github/actions/install-boost/action.yml
+++ b/.github/actions/install-boost/action.yml
@@ -32,7 +32,7 @@ runs:
       shell: bash
       run: |
         choco install boost-msvc-14.3 -y
-        echo "BOOST_ROOT=C:\local\boost_1_81_0" >> $GITHUB_OUTPUT
+        echo "BOOST_ROOT=C:\local\boost_1_82_0" >> $GITHUB_OUTPUT
 
     - name: Install boost using homebrew
       id: brew-action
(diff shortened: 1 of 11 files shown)


.github/actions/install-boost/action.yml
```diff
@@ -39,7 +39,7 @@ runs:
       if: runner.os == 'macOS'
       shell: bash
       run: |
-        brew install boost@1.81
+        brew install boost
```
kinyoklion, line 42: Unpinning this as it seems brew doesn't keep the previous packages as 1.8 is being updated.

.github/actions/install-boost/action.yml
```diff
@@ -32,15 +32,15 @@ runs:
       shell: bash
       run: |
         choco install boost-msvc-14.3 -y
-        echo "BOOST_ROOT=C:\local\boost_1_81_0" >> $GITHUB_OUTPUT
+        echo "BOOST_ROOT=C:\local\boost_1_82_0" >> $GITHUB_OUTPUT
```
kinyoklion, line 35: Ninja floated to 1.82 as well.

cwaldren-ld (approved):

PR #159, opened 2023-06-30
feat: add Flag model to SDKDataSet

cwaldren-ld:
Adds in missing `Flag` model to the top-level `SDKDataSet`.

I've also changed both the flag/segment fields to use `PARSE_FIELD`, so we can get an empty `unordered_map` if they are missing.

### DIFF
--- a/libs/internal/include/launchdarkly/data_model/flag.hpp
+++ b/libs/internal/include/launchdarkly/data_model/flag.hpp
@@ -84,6 +84,11 @@ struct Flag {
     bool trackEventsFallthrough;
     std::optional<std::uint64_t> debugEventsUntilDate;
 
+    /**
+     * Returns the flag's version. Satisfies ItemDescriptor template
+     * constraints.
+     * @return Version of this flag.
+     */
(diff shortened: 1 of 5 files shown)


libs/internal/include/launchdarkly/data_model/flag.hpp
```diff
@@ -83,5 +83,7 @@ struct Flag {
     bool trackEvents;
     bool trackEventsFallthrough;
     std::optional<std::uint64_t> debugEventsUntilDate;
+
+    [[nodiscard]] inline std::uint64_t Version() const { return version; }
```
kinyoklion, line 87: Confused.

kinyoklion (approved):

libs/internal/include/launchdarkly/data_model/flag.hpp
cwaldren-ld: Sorry, this deserved documentation - it's satisfying the `ItemDescriptor` template requirements. Will add a comment.

PR #160, opened 2023-07-06
feat: server-side SDK 

cwaldren-ld:
This PR exists to verify the linting, compilation, and test CI steps for the work-in-progress Server-Side SDK.

This feature branch does not yet represent the finished product but rather an alpha. Once merged, additional work is needed to onboard to release please.


.github/workflows/common.yml
```diff
@@ -6,12 +6,12 @@ on:
     paths-ignore:
       - '**.md' #Do not need to run CI for markdown changes.
   pull_request:
-    branches: [ main ]
```
cwaldren-ld, line 9: Can remove `server-side` once merged.

libs/internal/src/events/asio_event_processor.cpp
```diff
@@ -205,60 +209,93 @@ std::optional<EventBatch> AsioEventProcessor<SDK>::CreateBatch() {
...
+                if constexpr (std::is_same<SDK,
+                                           config::shared::ServerSDK>::value) {
+                    if (!context_key_cache_.Notice(
+                            event.context.CanonicalKey())) {
+                        out.emplace_back(server_side::IndexEvent{
+                            event.creation_date,
+                            filter_.filter(event.context)});
+                    }
+                }
+
```
cwaldren-ld, line 287: TODO: Could probably achieve this with a constructor instead.

libs/client-sdk/README.md
```diff
@@ -8,6 +8,8 @@ The LaunchDarkly Client-Side SDK for C/C++ is designed primarily for use in desk
 It follows the client-side LaunchDarkly model for single-user contexts (much like our mobile or JavaScript SDKs). It is
 not intended for use in multi-user systems such as web servers and applications.
```
cwaldren-ld, line 10: TODO: The readme currently lists the cmake variables used to control the build. Should we move this into the top-level README, or copy into the server-sdk's README?


 Right now they are all general rather than sdk-specific.

libs/internal/src/serialization/json_flag.cpp
```diff
@@ -0,0 +1,371 @@
...
+    obj.emplace("weight", weighted_variation.weight);
+
+    WriteMinimal(obj, "untracked", weighted_variation.untracked);
+}
+
+void tag_invoke(boost::json::value_from_tag const& unused,
+                boost::json::value& json_value,
+                data_model::Flag::Rollout::Kind const& kind) {
+    switch (kind) {
+        case Flag::Rollout::Kind::kUnrecognized:
```
cwaldren-ld, line 230: TODO

libs/internal/src/serialization/json_rule_clause.cpp
```diff
@@ -0,0 +1,178 @@
...
+        obj.emplace("attribute", clause.attribute.RedactionName());
+    }
+    obj.emplace("contextKind", clause.contextKind.t);
+}
+
+void tag_invoke(boost::json::value_from_tag const& unused,
+                boost::json::value& json_value,
+                data_model::Clause::Op const& op) {
+    switch (op) {
+        case data_model::Clause::Op::kUnrecognized:
```
cwaldren-ld, line 126: TODOs

libs/server-sdk/src/all_flags_state/json_all_flags_state.cpp
```diff
@@ -0,0 +1,56 @@
...
+    boost::ignore_unused(unused);
+    auto& obj = json_value.emplace_object();
+
+    if (!state.OmitDetails()) {
+        obj.emplace("version", state.Version());
+
+        if (auto const& reason = state.Reason()) {
+            obj.emplace("reason", boost::json::value_from(*reason));
+        }
+    }
```
cwaldren-ld, line 22: TODO: switching these to WriteMinimal

libs/server-sdk/src/client_impl.cpp
```diff
@@ -0,0 +1,459 @@
...
+        return std::make_shared<data_sources::NullDataSource>(executor,
+                                                              status_manager);
+    }
+
+    auto builder = HttpPropertiesBuilder(http_properties);
+
+    auto data_source_properties = builder.Build();
+
+    if (config.DataSourceConfig().method.index() == 0) {
+        // TODO: use initial reconnect delay.
```
cwaldren-ld, line 54: TODO

libs/internal/src/serialization/json_flag.cpp
cwaldren-ld: Thinking about this, it feels like the long term solution shouldn't rely on "SDK deserializes data, then reserializes for persistence based on its in-memory model". Otherwise, old SDKs that aren't aware of new fields will drop them and "pollute" or otherwise downgrade the data. Instead, they should pass-through to the storage - save the CPU cycles and allow newer clients (including newer versions of the same SDK) to read the complete data.

kinyoklion: I agree with this conceptually, but it has a couple problems which I tried to reason through when we started the V2 process. We need to update individual records and we likely do not want to store or re-create the textual record in order to update. (double memory, or base storage in a non-typed form, like keeping the boost::json representation).

We get an init, we save that to the persistent store. 

We get a patch for a flag in that init, what do we do? Read back the json, deserialize it, do patch comparisons, update it? Ok, now how do we update the `allFlags` key? We cannot really patch JSON without putting into some representable form, changing it, and then serializing it. Basically we have parallel implementations of the patch update process.

I think that we possibly could do something like this, but it would require changing how we store the representation. I am not sure how yet. Certainly dropping the idea of the possibility of a store without the ability to scan, which is where the possibility of an 'all' key comes from.

kinyoklion: So, "all must be an aggregate of all the entries for the specified type". Which would hold true for dynamodb and redis.

libs/client-sdk/include/launchdarkly/client_side/data_source_status.hpp
```diff
@@ -5,202 +5,75 @@
...
 #include <launchdarkly/connection.hpp>
+#include <launchdarkly/data_sources/data_source_status_base.hpp>
 
 namespace launchdarkly::client_side::data_sources {
 
-class DataSourceStatus {
-   public:
-    using DateTime = std::chrono::time_point<std::chrono::system_clock>;
-
+/**
```
cwaldren-ld, line 15: Calling out as a theoretical breaking change.

 In previous Client SDK, you could call `IDataSourceStatusProvider::Status()` and get back a `class client_side::data_sources::DataSourceStatus`.

Now, you get back a `class common::data_sources::DataSourceStatusBase<DataSourceState>`. 

These classes should have the same API, but the concrete type has changed. 

If the caller did:
```c++
auto status = sdk.DataSourceStatus().Status();
if (auto err = status.LastError()) {
   ...
}
```
then this should compile without any changes. But if they did:
```
client_side::data_sources::DataSourceStatus status = sdk.DataSourceStatus().Status();
```
that's going to require a change.

We have [some](https://github.com/launchdarkly/cpp-sdks/releases/tag/launchdarkly-cpp-client-v3.0.6) [precedent ](https://github.com/launchdarkly/cpp-sdks/releases/tag/launchdarkly-cpp-client-v3.0.5)for "non-major to fix a bug in the API definition".

cwaldren-ld:
I've reviewed all files, fixed a bunch of outstanding TODOs, and got this branch all caught up with main.

I'm feeling ready to merge pending a backwards compatibility concern I've called out in comments.

libs/client-sdk/include/launchdarkly/client_side/data_source_status.hpp
cwaldren-ld: Actually, this might be fine. Since `client_side::data_sources::DataSourceStatus` is a type alias, and the new type didn't exist, it should still compile. It's definitely ABI breaking but we don't claim to support a stable C++ ABI.

cwaldren-ld: The following code compiles without issue on the existing client, and the client in this PR:
```c++
  client_side::data_sources::DataSourceStatus ds_status =
        client.DataSourceStatus().Status();

    std::optional<client_side::data_sources::DataSourceStatus::ErrorInfo>
        last_err = ds_status.LastError();

    client_side::data_sources::DataSourceStatus::DataSourceState state =
        ds_status.State();

    client_side::data_sources::DataSourceStatus::DateTime date =
        ds_status.StateSince();
```

libs/internal/src/serialization/json_flag.cpp
cwaldren-ld: Added a todo link into the codebase for this. Deserves a more extended discussion, for sure.

kinyoklion (approved):

PR #165, opened 2023-07-10
feat: Implement basic in-memory store and change handling.

kinyoklion:
feat: Move boost signal implementation to the internal library.


libs/client-sdk/src/CMakeLists.txt
```diff
@@ -17,10 +17,8 @@ add_library(${LIBNAME}
         data_sources/data_source_status_manager.cpp
         event_processor/event_processor.cpp
         event_processor/null_event_processor.cpp
-        boost_signal_connection.cpp
```
kinyoklion, line 20: Moved boost_signal_connection to internal.

libs/server-sdk/src/data_store/dependency_tracker.hpp
```diff
@@ -0,0 +1,165 @@
...
+   private:
+    DataType data_;
+};
+
+/**
+ * This class implements a mechanism of tracking dependencies of flags and
+ * segments. Both the forward dependencies (flag A depends on flag B) but also
+ * the reverse (flag B is depended on by flagA).
+ */
+class DependencyTracker {
```
kinyoklion, line 105: This dependency tracker works basically the same as the one in node and dotnet.

kinyoklion: There are lots of strings in this representation. It is probably fine, but something to consider if we have any performance issues.

libs/server-sdk/src/data_store/data_store_updater.hpp
```diff
@@ -0,0 +1,93 @@
...
+
+#include <launchdarkly/server_side/change_notifier.hpp>
+
+#include <boost/signals2/signal.hpp>
+
+#include <memory>
+
+namespace launchdarkly::server_side::data_store {
+
+class DataStoreUpdater
```
kinyoklion, line 15: I wonder why my local linter doesn't complain about this. Time to default all the things!

kinyoklion:
Converted to draft because I missed some testing and found a couple issues.

libs/server-sdk/src/data_store/dependency_tracker.cpp
```diff
@@ -0,0 +1,186 @@
...
+
+void DependencySet::Set(DataKind kind, std::string key) {
+    data_[static_cast<std::size_t>(kind)].Data().emplace(std::move(key));
+}
+
+void DependencySet::Remove(DataKind kind, std::string const& key) {
+    data_[static_cast<std::size_t>(kind)].Data().erase(key);
+}
+
+bool DependencySet::Contains(DataKind kind, std::string const& key) {
```
cwaldren-ld, line 19: Should be const?

libs/server-sdk/src/data_store/dependency_tracker.hpp
```diff
@@ -0,0 +1,165 @@
...
+ * is for. This is primarily to decrease the complexity of iterating collections
+ * allowing for a kvp style iteration, but with an array storage container.
+ * @tparam Storage
+ */
+template <typename Storage>
+class TaggedData {
+   public:
+    explicit TaggedData(DataKind kind) : kind_(kind) {}
+    [[nodiscard]] DataKind Kind() const { return kind_; }
+    [[nodiscard]] Storage& Data() { return storage_; }
```
cwaldren-ld, line 26: Could add a `Storage const& Data() const` accessor (I think would be needed if you were to make `DependencySet::Contains` const)

libs/server-sdk/src/data_store/dependency_tracker.hpp
```diff
@@ -0,0 +1,165 @@
...
+
+    [[nodiscard]] bool Contains(DataKind kind, std::string const& key);
+
+    [[nodiscard]] std::set<std::string> SetForKind(DataKind kind);
+
+    /**
+     * Return the size of all the data kind sets.
+     * @return The combined size of all the data kind sets.
+     */
+    [[nodiscard]] std::size_t Size();
```
cwaldren-ld, line 59: const?

libs/server-sdk/src/data_store/data_store_updater.cpp
```diff
@@ -0,0 +1,78 @@
...
+    dependency_tracker_.Clear();
+    for (auto const& flag : data_set.flags) {
+        dependency_tracker_.UpdateDependencies(flag.first, flag.second);
+    }
+    for (auto const& segment : data_set.segments) {
+        dependency_tracker_.UpdateDependencies(segment.first, segment.second);
+    }
+    // Data will move into the store, so we want to update dependencies before
+    // it is moved.
+    sink_->Init(data_set);
```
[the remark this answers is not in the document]
kinyoklion: There should be a move here as well, but the comment refers to the actual descriptors.
```std::make_shared<IDataStore::FlagDescriptor>(std::move(flag));```
In that we move the flag out of this descriptor and into the descriptor in the actual store.

File: libs/server-sdk/src/data_store/data_store_updater.cpp [hunk 8]
```diff
@@ -0,0 +1,78 @@
...
+    // After updating the sink let listeners know of changes.
+    if (change_notifications) {
+        NotifyChanges(*change_notifications);
+    }
+}
+
+void DataStoreUpdater::Upsert(std::string key,
+                              launchdarkly::server_side::data_source::
+                                  IDataSourceUpdateSink::FlagDescriptor flag) {
+    UpsertCommon(DataKind::kFlag, key, store_->GetFlag(key), flag);
```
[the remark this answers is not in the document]
kinyoklion: That was the plan, but I think it got lost in the layers.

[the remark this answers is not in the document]
kinyoklion: I may just make more things const and let it copy, otherwise it is really easy to use the key twice in one function and it to be invalid. Downside of the way C++ does moves.

libs/server-sdk/src/data_store/data_store_updater.hpp
kinyoklion: So, the thing that was here, where the linter was just exploding, is gone now. Which is maybe why I got other new lints.

libs/server-sdk/src/data_store/dependency_tracker.cpp
kinyoklion: I made it non-const because making it const was hell. 
Because then `.Data()` needs a const equivalent. Turns into rust.

kinyoklion: I can see if I can make it less terrible.

libs/server-sdk/src/data_store/dependency_tracker.cpp
```diff
@@ -0,0 +1,186 @@
...
+// NOLINTBEGIN misc-no-recursion
+
+void DependencyTracker::CalculateChanges(DataKind kind,
+                                         std::string const& key,
+                                         DependencySet& dependency_set) {
+    if (!dependency_set.Contains(kind, key)) {
+        dependency_set.Set(kind, key);
+        auto affected_items = dependencies_to_.Get(kind, key);
+        if (affected_items) {
+            for (auto& depNs : *affected_items) {
```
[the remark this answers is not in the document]
kinyoklion: I mean't to rename this.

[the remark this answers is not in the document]
kinyoklion: (It was the dependency namespace, when lots of other things were different.)

libs/server-sdk/src/data_store/dependency_tracker.cpp
```diff
@@ -0,0 +1,186 @@
...
+
+// NOLINTEND misc-no-recursion
+
+void DependencyTracker::UpdateDependencies(DataKind kind,
+                                           std::string const& key,
+                                           DependencySet deps) {
+    auto current_deps = dependencies_from_.Get(kind, key);
+    if (current_deps) {
+        for (auto& dep_kind : *current_deps) {
+            auto depKind = dep_kind.Kind();
```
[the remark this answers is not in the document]
kinyoklion: All the names that didn't get case converted are wrong.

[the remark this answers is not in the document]
kinyoklion: Though this one will still be kind of confusing. Because it is hard to have "all of the items of kind, and the kind", and "just the kind".

libs/server-sdk/src/data_store/data_store.hpp
```diff
@@ -0,0 +1,85 @@
...
+#include <memory>
+#include <string>
+#include <unordered_map>
+
+namespace launchdarkly::server_side::data_store {
+
+/**
+ * Interface for readonly access to SDK data.
+ *
+ * The data store is what the client uses to store feature flag data that has
```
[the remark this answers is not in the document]
kinyoklion: It says what it does, and is accurate. The IDataStore isn't directly the data store. 

That said, I just deleted it.

libs/server-sdk/src/data_store/data_store_updater.hpp
```diff
@@ -0,0 +1,122 @@
...
+        if (HasListeners()) {
+            auto updated_deps = DependencySet();
+            dependency_tracker_.CalculateChanges(kind, key, updated_deps);
+            NotifyChanges(updated_deps);
+        }
+
+        sink_->Upsert(key, updated);
+    }
+
+    template <typename FlagOrSegment>
```
kinyoklion, line 69: Still FlagOrSegment, because DataKind is a type, which is also used by this same method. Which is inconvenient.

cwaldren-ld (approved):
Couple lints on `libs/server-sdk/src/data_store/dependency_tracker.cpp` seem legit, but otherwise looks good!

PR #166, opened 2023-07-11
feat: evaluation algorithm

cwaldren-ld:
This PR contains the main evaluation engine and associated tests.


libs/common/include/launchdarkly/data/evaluation_detail.hpp
```diff
@@ -52,21 +60,33 @@ class EvaluationDetail {
...
      * @return A reference to the variation value.
      */
     T const& operator*() const;
 
+    /**
+     * @return True if the evaluation was successful (i.e. IsError returns
+     * false.)
+     */
+    explicit operator bool() const;
+
```
cwaldren-ld, line 78: Unclear to me if we need both `IsError` and `operator bool()`. 

The intention is to allow users (and us, in test code) to check if the evaluation has an error, without having to reach deep into the object. 

Having `if (detail)` is pretty convenient and fits in with `std::optional` and `tl::expected`, I just don't know if we need the explicit `IsError` as well.

kinyoklion: std::optional, and tl::expected both have kind of pointer like behavior, and they both additionally have an explicit way to check for a value `has_value`. We don't really have pointer like behavior, so I don't know that the boolean operator would be very apparent. It isn't bad to have, but an error and having a value are a bit different for an evaluation reason.

libs/internal/include/launchdarkly/logging/spy_logger.hpp
```diff
@@ -0,0 +1,34 @@
...
+ * Creates a logger for tests where log messages need to be inspected.
+ *
+ * @return Logger suitable for passing to functions accepting a
+ * launchdarkly::Logger.
+ */
+launchdarkly::Logger SpyLogger(std::vector<std::string>& messages);
+
+class SpyLoggerBackend : public launchdarkly::ILogBackend {
+   public:
+    SpyLoggerBackend(std::vector<std::string>& messages);
```
kinyoklion, line 19: I think the vector should be tuples containing the level and the string value. So you can validate not just the log message, but that it was logged as the correct level.

suggestion: A helper for asserting on one or more expected messages lowers the barrier of entry quite a bit as well. 
Example from js-core to illustrate.
```node
      logger.expectMessages([
        {
          level: LogLevel.Error,
          matches: /Received error.*giving up permanently/,
        },
```

libs/common/include/launchdarkly/data/evaluation_detail.hpp
cwaldren-ld: It's true it's not pointer-like, but we do have `operator*` to get at the value. So you could do:
```c++
 if (result) {
      double x = (*result).AsDouble()
 }
```
But this does feel clunky. Hmm.

cwaldren-ld: Perhaps the `VariationDetail` methods should really accept detail as an out-param.

libs/internal/include/launchdarkly/logging/spy_logger.hpp
cwaldren-ld: That's fair. I'll add that.

libs/common/include/launchdarkly/data/evaluation_detail.hpp
kinyoklion: Generally speaking I don't think you usually even care if there was an error.

libs/server-sdk/src/evaluation/rules.cpp
```diff
@@ -0,0 +1,215 @@
...
+
+    if (IsTargeted(context, segment.excluded, segment.excludedContexts)) {
+        return false;
+    }
+
+    for (auto const& rule : segment.rules) {
+        if (!segment.salt) {
+            return tl::make_unexpected(Error::MissingSalt(segment.key));
+        }
+        if (Match(rule, context, store, stack, segment.key, *segment.salt)) {
```
kinyoklion, line 180: Is this correct? Reflexively it feels incorrect, but maybe I am not following the logic correctly (I would expect the value to be used, versus presence of value?).

kinyoklion (changes requested):
Inline question about segment matching.

libs/internal/include/launchdarkly/logging/spy_logger.hpp
cwaldren-ld: Added this API using `:::testing::AssertionResult` helpers:
```c++
    ASSERT_TRUE(messages_->Count(1));
    ASSERT_TRUE(messages_->Contains(0, LogLevel::kError, "some string"));
```
Can extend if we need more matchers.

kinyoklion: I like it.

libs/server-sdk/src/evaluation/rules.cpp
cwaldren-ld: Yes, this is wrong. I missed that this was returning a `tl::expected`. Updated and added a unit test to make sure this piece is getting exercised.

kinyoklion (approved):

PR #172, opened 2023-07-12
feat: add SemVer class and associated parsing routine

cwaldren-ld:
Adds a `SemVer` class to perform semantic version parsing. 

I implemented semver operators natively instead of pulling a 3rd party dependency like C Server mainly because string manipulation is tractable in C++.

Additionally the C Server needed to patch the dependency to support LD-flavored semver; no patches needed here.

### DIFF
--- a/libs/server-sdk/src/evaluation/detail/semver_operations.cpp
+++ b/libs/server-sdk/src/evaluation/detail/semver_operations.cpp
@@ -1,0 +1,173 @@
+#include "semver_operations.hpp"
+
+#include <launchdarkly/detail/c_binding_helpers.hpp>
+
+#include <boost/algorithm/string.hpp>
+#include <boost/regex.hpp>
+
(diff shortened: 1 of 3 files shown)


libs/server-sdk/src/evaluation/detail/semver_operations.cpp
```diff
@@ -0,0 +1,203 @@
...
+        return false;
+    }
+    return *lhs.Prerelease() < *rhs.Prerelease();
+}
+
+bool operator>(SemVer const& lhs, SemVer const& rhs) {
+    return rhs < lhs;
+}
+
+std::optional<SemVer> SemVer::Parse(std::string const& value) {
```
cwaldren-ld, line 116: using `boost::lexical_cast` instead of `std::stoull` because the latter accepts things like "2-3" as the number 2.

kinyoklion (approved):

PR #175, opened 2023-07-12
feat: context bucketing logic

cwaldren-ld:
Adds bucketing logic to be used in evaluation engine. 

Buckets are computed based on a hash value comprised of context attributes and prefixes derived from flag rules.

### DIFF
--- a/libs/server-sdk/src/evaluation/bucketing.cpp
+++ b/libs/server-sdk/src/evaluation/bucketing.cpp
@@ -77,6 +77,8 @@ AttributeReference const& Key() {
 }
 
 std::optional<float> ContextHash(Value const& value, BucketPrefix prefix) {
+    using namespace launchdarkly::encoding;
+
     std::optional<std::string> id = BucketValue(value);
     if (!id) {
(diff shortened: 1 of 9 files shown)


libs/server-sdk/src/evaluation/bucketing.cpp
```diff
@@ -0,0 +1,130 @@
...
+                os << arg;
+            }
+        },
+        prefix.prefix_);
+    return os;
+}
+
+tl::expected<std::pair<ContextHashValue, RolloutKindLookup>, Error> Bucket(
+    Context const& context,
+    AttributeReference const& by_attr,
```
kinyoklion, line 44: I assume this is handled elsewhere, maybe in parsing, but when bucketBy is omitted entirely it is also "key".

kinyoklion (approved):

libs/server-sdk/src/evaluation/bucketing.cpp
cwaldren-ld: Yep, it's in the parser.

PR #176, opened 2023-07-12
feat: initial pass of server-side Client object

cwaldren-ld:
This is a first pass at creating the server-side Client object / facade. It is based on the client-side object of the same name.

I've commented out some parts to get it building and market them with TODOs.


libs/client-sdk/src/CMakeLists.txt
```diff
@@ -24,8 +22,6 @@ add_library(${LIBNAME}
         data_sources/data_source_update_sink.hpp
         data_sources/polling_data_source.hpp
         data_sources/streaming_data_source.hpp
-        event_processor/event_processor.hpp
-        event_processor/null_event_processor.hpp
         flag_manager/flag_store.hpp
```
cwaldren-ld, line 25: These live in `internal` now, so they can be shared with the server-side SDK.

libs/client-sdk/src/client_impl.cpp
```diff
@@ -112,14 +110,15 @@ ClientImpl::ClientImpl(Config config,
     flag_manager_.LoadCache(context_);
 
     if (config.Events().Enabled() && !config.Offline()) {
-        event_processor_ = std::make_unique<EventProcessor>(
-            ioc_.get_executor(), config.ServiceEndpoints(), config.Events(),
-            http_properties_, logger_);
+        event_processor_ =
+            std::make_unique<events::AsioEventProcessor<ClientSDK>>(
+                ioc_.get_executor(), config.ServiceEndpoints(), config.Events(),
+                http_properties_, logger_);
```
cwaldren-ld, line 116: I've removed a now-redundant layer here. Before, `EventProcessor` was a very thin wrapper that forwarded its arguments to `AsioEventProcessor`. This worked fine when only the client SDK existed. Ideally, we could share that interface with the server.

Now that `AsioEventProcessor` implements `IEventProcessor`, the server and client SDKs can directly instantiate it instead.

libs/internal/include/launchdarkly/events/client_events.hpp
```diff
@@ -1,50 +0,0 @@
-#pragma once
-
```
cwaldren-ld, line 2: These are no longer client-specific, so I've moved them to a `common_events.hpp`.

libs/server-sdk/src/client_impl.cpp
```diff
@@ -0,0 +1,373 @@
...
+    event_processor_->FlushAsync();
+}
+
+template <typename T>
+EvaluationDetail<T> ClientImpl::VariationInternal(Context const& ctx,
+                                                  FlagKey const& key,
+                                                  Value default_value,
+                                                  bool check_type,
+                                                  bool detailed) {
+    //    auto desc = memory_store_.GetFlag(key);
```
cwaldren-ld, line 203: Evaluator PR isn't merged yet.

libs/server-sdk/include/launchdarkly/server_side/client.hpp
```diff
@@ -0,0 +1,331 @@
...
+#include <memory>
+#include <string>
+#include <unordered_map>
+
+namespace launchdarkly::server_side {
+
+/**
+ *  Interface for the standard SDK client methods and properties.
+ */
+class IClient {
```
cwaldren-ld, line 19: This whole interface is copied from the client, so adjustments will need to be made in subsequent PRs.

libs/server-sdk/include/launchdarkly/server_side/data_source_status.hpp
```diff
@@ -61,6 +61,12 @@ enum class ServerDataSourceState {
      * explicitly closed.
      */
     kOff = 3,
+
+    /**
+     * Indicates that the application has told the SDK to stay offline.
+     */
+    kSetOffline = 4,
```
cwaldren-ld, line 68: Wasn't sure if `kOff` was appropriate/intended for offline mode.

libs/internal/tests/event_serialization_test.cpp
```diff
@@ -3,7 +3,7 @@
 #include <boost/json.hpp>
 
 #include <launchdarkly/context_builder.hpp>
-#include <launchdarkly/events/events.hpp>
+#include "launchdarkly/events/data/events.hpp"
```
kinyoklion, line 6: `<>`=>`""`?

libs/server-sdk/include/launchdarkly/server_side/data_source_status.hpp
kinyoklion: `kSetOffline` is a client side data source state, but doesn't seem to be a server side one. I think it would be kOff. 

(Go and dotnet only have the states I already defined here, but I don't really see an actual spec.) 

Neither of them say, in the documentation, which it is.

That said, even for client, it shouldn't be set offline. I think SetOffline is only that you dynamically made it offline. Where offline in the config is shutdown.

kinyoklion (approved):

kinyoklion (changes requested):
Misclick. Meant to request changes for the data source state.

PR #179, opened 2023-07-13
feat: Implement streaming data source.

kinyoklion:


libs/server-sdk/src/data_sources/data_source_event_handler.cpp
```diff
@@ -0,0 +1,218 @@
...
+    }
+    auto data =
+        boost::json::value_to<tl::expected<std::optional<TData>, JsonError>>(
+            data_iter->value());
+    if (!data.has_value()) {
+        return tl::unexpected(JsonError::kSchemaFailure);
+    }
+    return DataSourceEventHandler::Patch{
+        TStreamingDataKind::Key(path),
+        data_model::ItemDescriptor<TData>(data->value())};
```
kinyoklion, line 46: I wish this lint worked, because it would be useful, but it is basically always wrong. Checked on line 41.

libs/server-sdk/src/data_sources/data_source_event_handler.cpp
```diff
@@ -0,0 +1,218 @@
...
+        tl::expected<DataSourceEventHandler::Put, JsonError>> const& unused,
+    boost::json::value const& json_value) {
+    boost::ignore_unused(unused);
+
+    if (!json_value.is_object()) {
+        return tl::unexpected(JsonError::kSchemaFailure);
+    }
+
+    DataSourceEventHandler::Put put;
+    PARSE_FIELD(put.data, json_value.as_object(), "data");
```
kinyoklion, line 60: This lint is in the macro? I thought we had exclusions for it.

libs/server-sdk/src/data_sources/data_source_status.cpp
```diff
@@ -0,0 +1,39 @@
...
+    return out;
+}
+
+std::ostream& operator<<(std::ostream& out, DataSourceStatus const& status) {
+    std::time_t as_time_t =
+        std::chrono::system_clock::to_time_t(status.StateSince());
+    out << "Status(" << status.State() << ", Since("
+        << std::put_time(std::gmtime(&as_time_t), "%Y-%m-%d %H:%M:%S") << ")";
+    if (status.LastError().has_value()) {
+        out << ", " << status.LastError().value();
```
kinyoklion, line 33: I think I understand why this is offensive to the linter, technically LastError *could* return a different optional.

libs/server-sdk/src/data_sources/streaming_data_source.cpp
```diff
@@ -0,0 +1,152 @@
...
+    if (!uri_components) {
+        LD_LOG(logger_, LogLevel::kError) << kCouldNotParseEndpoint;
+        status_manager_.SetState(
+            DataSourceStatus::DataSourceState::kOff,
+            DataSourceStatus::ErrorInfo::ErrorKind::kNetworkError,
+            kCouldNotParseEndpoint);
+        return;
+    }
+
+    // TODO: Initial reconnect delay. sc-204393
```
kinyoklion, line 78: Likely this should still be addressed as an independent change, but maybe this ticket should be pulled into the epic.
I think that the SSE client currently has a hardcoded value and it just needs to expose a value for building instead, and then we can use that value.

libs/server-sdk/src/data_sources/streaming_data_source.cpp
```diff
@@ -0,0 +1,152 @@
...
+
+    // TODO: Handle proxy support. sc-204386
+
+    auto weak_self = weak_from_this();
+
+    client_builder.receiver([weak_self](launchdarkly::sse::Event const& event) {
+        if (auto self = weak_self.lock()) {
+            self->data_source_handler_.HandleMessage(event.type(),
+                                                     event.data());
+            // TODO: Use the result of handle message to restart the
```
kinyoklion, line 106: I am not sure how to handle this in the current pattern. I think we went with re-creating instead of restarting. So it will require extracting out the build of the client, then closing and re-creating the client.

libs/server-sdk/src/data_sources/polling_data_source.cpp
```diff
@@ -0,0 +1,245 @@
...
+#include "polling_data_source.hpp"
+
+namespace launchdarkly::server_side::data_sources {
+
+static char const* const kErrorParsingPut = "Could not parse polling payload";
+static char const* const kErrorPutInvalid =
+    "polling payload contained invalid data";
+
+static char const* const kCouldNotParseEndpoint =
+    "Could not parse polling endpoint URL.";
```
cwaldren-ld, line 23:
```suggestion
    "Could not parse polling endpoint URL";
```

libs/server-sdk/src/data_sources/polling_data_source.cpp
```diff
@@ -0,0 +1,245 @@
...
+    if (res.IsError()) {
+        auto const& error_message = res.ErrorMessage();
+        status_manager_.SetState(
+            DataSourceStatus::DataSourceState::kInterrupted,
+            DataSourceStatus::ErrorInfo::ErrorKind::kNetworkError,
+            error_message.has_value() ? *error_message : "unknown error");
+        LD_LOG(logger_, LogLevel::kWarn)
+            << "Polling for feature flag updates failed: "
+            << (error_message.has_value() ? *error_message : "unknown error");
+    } else if (res.Status() == 200) {
```
cwaldren-ld, line 130: Kinda wish we could use the beast error codes here. Maybe we can, they might just be the numeric values of the status codes or have a getter for it.

libs/server-sdk/src/data_sources/streaming_data_source.cpp
```diff
@@ -0,0 +1,152 @@
...
+#include <utility>
+
+#include "streaming_data_source.hpp"
+
+#include <launchdarkly/network/http_requester.hpp>
+
+namespace launchdarkly::server_side::data_sources {
+
+static char const* const kCouldNotParseEndpoint =
+    "Could not parse streaming endpoint URL.";
```
cwaldren-ld, line 16:
```suggestion
    "Could not parse streaming endpoint URL";
```
Just commenting for consistency with the other logs. But we could also chose to terminate them all with periods.

libs/server-sdk/src/data_sources/streaming_data_source.cpp
cwaldren-ld: Another thought is have the callback return a bool/status that tells the eventsource to reconnect. Then we don't have to rebuild it and can keep the async stuff hidden there.

libs/server-sdk/src/data_sources/data_source_event_handler.cpp
```diff
@@ -0,0 +1,220 @@
...
+        return tl::unexpected(JsonError::kSchemaFailure);
+    }
+
+    auto const& obj = json_value.as_object();
+
+    std::string path;
+    PARSE_FIELD(path, obj, "path");
+
+    if (StreamingDataKinds::Flag::IsKind(path)) {
+        return Patch<StreamingDataKinds::Flag, data_model::Flag>(path, obj);
```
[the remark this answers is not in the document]
kinyoklion: I spent an inordinate amount of time on template shenanigans. Ultimately it is pretty pleasant though,

libs/server-sdk/src/data_sources/data_source_event_handler.cpp
```diff
@@ -0,0 +1,220 @@
...
+    }
+
+    if (StreamingDataKinds::Segment::IsKind(path)) {
+        return Patch<StreamingDataKinds::Segment, data_model::Segment>(path,
+                                                                       obj);
+    }
+
+    return tl::unexpected(JsonError::kSchemaFailure);
```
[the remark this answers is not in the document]
kinyoklion: I may have this wrong.

libs/server-sdk/src/data_sources/streaming_data_source.cpp
kinyoklion: That sounds more pleasant.

libs/server-sdk/src/data_sources/data_source_event_handler.cpp
```diff
@@ -0,0 +1,242 @@
+#include "data_source_event_handler.hpp"
+
+#include <launchdarkly/encoding/base_64.hpp>
+#include <launchdarkly/serialization/json_flag.hpp>
+// #include <launchdarkly/serialization/json_primitives.hpp>
```
cwaldren-ld, line 5: remove

cwaldren-ld (approved):

PR #184, opened 2023-07-18
feat: add ContextKind type to data model

cwaldren-ld:
Previously, context kind was an `std::string`. 

The requirement that it not be an empty string was enforced by parent data model deserializers. 

Now, we have a dedicated type-safe wrapper via`BOOST_STRONG_TYPEDEF`. The custom `tag_invoke` enforces the invariant so it's not possible to deserialize it inconsistently.

### DIFF
--- a/libs/internal/include/launchdarkly/data_model/context_kind.hpp
+++ b/libs/internal/include/launchdarkly/data_model/context_kind.hpp
@@ -1,0 +1,11 @@
+#pragma once
+
+#include <boost/serialization/strong_typedef.hpp>
+
+#include <string>
+
+namespace launchdarkly::data_model {
+
+BOOST_STRONG_TYPEDEF(std::string, ContextKind);
+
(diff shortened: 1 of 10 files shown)


libs/internal/include/launchdarkly/data_model/context_kind.hpp
```diff
@@ -0,0 +1,11 @@
+#pragma once
+
+#include <boost/serialization/strong_typedef.hpp>
+
+#include <string>
+
+namespace launchdarkly::data_model {
+
+BOOST_STRONG_TYPEDEF(std::string, ContextKind);
```
cwaldren-ld, line 9: It is very nice that this exists.

kinyoklion: I suppose I could have used this in place of my tag types for flags/segments.

kinyoklion (approved):

PR #186, opened 2023-07-18
fix: avoid copying Value in Evaluator algorithm

cwaldren-ld:
Removes potentially costly, unecessary copies of `Value` in the evaluation algorithm.

### DIFF
--- a/libs/server-sdk/src/evaluation/bucketing.cpp
+++ b/libs/server-sdk/src/evaluation/bucketing.cpp
@@ -73,7 +73,7 @@ tl::expected<std::pair<ContextHashValue, RolloutKindLookup>, Error> Bucket(
             Error::InvalidAttributeReference(ref.RedactionName()));
     }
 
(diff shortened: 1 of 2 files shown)


libs/server-sdk/src/evaluation/bucketing.cpp
```diff
@@ -73,7 +73,7 @@ tl::expected<std::pair<ContextHashValue, RolloutKindLookup>, Error> Bucket(
             Error::InvalidAttributeReference(ref.RedactionName()));
     }
 
-    Value value = context.Get(context_kind, ref);
+    Value const& value = context.Get(context_kind, ref);
```
cwaldren-ld, line 77: This is the more worrisome one.

kinyoklion: We may want to consider some temporary static inspection. Track how many values are copied.

kinyoklion (approved):

PR #187, opened 2023-07-19
feat: Add persistent store core interface.

kinyoklion:


libs/server-sdk/include/launchdarkly/server_side/integrations/persistent_store_core.hpp
```diff
@@ -0,0 +1,212 @@
+#pragma once
```
kinyoklion, line 1: Likely I will have to move this file to common in a subsequent PR. Because the config will need to contain this interface.

libs/server-sdk/include/launchdarkly/server_side/integrations/persistent_store_core.hpp
```diff
@@ -0,0 +1,212 @@
...
+     * std::nullopt for deleted items.
+     */
+    std::optional<std::string> serializedItem;
+};
+
+/**
+ * Represents a namespace of persistent data.
+ */
+class IPersistentKind {
+    /**
```
cwaldren-ld, line 38: needs a `public:` I think

kinyoklion: Indeed.

cwaldren-ld (approved):

PR #188, opened 2023-07-19
feat: Add expiration tracker.

kinyoklion:
This adds an class for tracking expiration of "scoped" (DataKind) and "unscoped" (initialized) keys.

Time handling is left to the caller, and proactive eviction could be handled using the prune methods.

(If we end up wanting to encapsulate time, then we could add a layer that just calls `steady_clock::now()`


libs/server-sdk/src/data_store/dependency_tracker.hpp
```diff
@@ -9,29 +9,10 @@
 #include <launchdarkly/data_model/segment.hpp>
 
 #include "data_kind.hpp"
+#include "tagged_data.hpp"
 
 namespace launchdarkly::server_side::data_store {
 
-/**
```
kinyoklion, line 15: Moved this to its own file so it could be re-used.

libs/server-sdk/src/data_store/persistent/persistent_data_store.hpp
```diff
@@ -0,0 +1,51 @@
+#pragma once
```
kinyoklion, line 1: I scaffolded this file, but have not started on implementation.

libs/server-sdk/src/data_store/tagged_data.hpp
```diff
@@ -0,0 +1,37 @@
+#pragma once
```
kinyoklion, line 1: Moved from dependency_tracker.hpp

libs/server-sdk/src/data_store/persistent/persistent_data_store.hpp
```diff
@@ -0,0 +1,51 @@
...
+    PersistentStore& operator=(PersistentStore const&) = delete;
+    PersistentStore& operator=(PersistentStore&&) = delete;
+
+   private:
+    MemoryStore memory_store_;
+    std::shared_ptr<integrations::IPersistentStoreCore> persistent_store_core_;
+    ExpirationTracker ttl_tracker_;
+};
+
+}  // namespace launchdarkly::server_side::data_store::persistent
```
kinyoklion, line 51: Adding a blank line.

libs/server-sdk/src/data_store/persistent/expiration_tracker.cpp
```diff
@@ -0,0 +1,152 @@
...
+    unscoped_.clear();
+}
+std::vector<std::pair<std::optional<DataKind>, std::string>>
+ExpirationTracker::Prune(ExpirationTracker::TimePoint current_time) {
+    std::vector<std::pair<std::optional<DataKind>, std::string>> pruned;
+
+    // Determine everything to be pruned.
+    for (auto const& item : unscoped_) {
+        if (State(item.second, current_time) ==
+            ExpirationTracker::TrackState::kStale) {
```
cwaldren-ld, line 58: I think both of these `push_back({arg1, arg2})` can be replaced with `emplace_back(arg1, arg2)` to avoid constructing a temporary

libs/server-sdk/src/data_store/persistent/expiration_tracker.hpp
```diff
@@ -0,0 +1,143 @@
...
+        ExpirationTracker::TimePoint expiration,
+        ExpirationTracker::TimePoint current_time);
+
+    class ScopedTtls {
+       public:
+        ScopedTtls();
+
+        using DataType =
+            std::array<TaggedData<TtlMap>,
+                       static_cast<std::size_t>(DataKind::kKindCount)>;
```
cwaldren-ld, line 122: I think use the `static_cast<std::underlying_type_t<`

cwaldren-ld (approved):

PR #190, opened 2023-07-19
fix: EvaluationStack should take ownership of key argument

cwaldren-ld:
Current API of `EvaluationStack` is very easy to accidentally misuse because it takes a reference to its key argument. If the arg is destroyed, then we'll have a use-after-free.

### DIFF
--- a/libs/server-sdk/src/evaluation/detail/evaluation_stack.cpp
+++ b/libs/server-sdk/src/evaluation/detail/evaluation_stack.cpp
@@ -2,7 +2,7 @@
 
 namespace launchdarkly::server_side::evaluation::detail {
 
-Guard::Guard(std::unordered_set<std::string>& set, std::string const& key)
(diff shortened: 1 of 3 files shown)


kinyoklion (approved):

libs/server-sdk/src/evaluation/detail/evaluation_stack.cpp
```diff
@@ -2,7 +2,7 @@
 
 namespace launchdarkly::server_side::evaluation::detail {
 
-Guard::Guard(std::unordered_set<std::string>& set, std::string const& key)
+Guard::Guard(std::unordered_set<std::string>& set, std::string key)
     : set_(set), key_(key) {
```
kinyoklion, line 6: Linter would like a move for the key.

kinyoklion (approved):

PR #192, opened 2023-07-20
chore: add event processor architectural diagrams

cwaldren-ld:
Also fixes some style / const-correct issues that I found while documenting.

### DIFF
--- a/architecture/event_processor.md
+++ b/architecture/event_processor.md
@@ -79,48 +79,49 @@ classDiagram
         +const EndTime() Time
     }
 
-    namespace events {
-        class InputEvent {
-            +std:: variant
-        }
-
-
-        class OutputEvent {
-            +std:: variant
-        }
-
-        class FeatureEventParams {
-
-        }
-
-        class IdentifyEventParams {
-
-        }
-
-        class TrackEventParams {
-
-        }
-
-        class FeatureEvent {
-
-        }
-
-        class DebugEvent {
-
-        }
-
-        class IdentifyEvent {
-
-        }
-
-        class IndexEvent {
-
-        }
-
-        class TrackEvent {
-
-        }
+%% note: the 'namespace' feature isn't supported on Github yet
+%% namespace events { 
+    class InputEvent {
+        +std:: variant
     }
+
+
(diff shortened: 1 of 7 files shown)


architecture/event_processor.md
```diff
@@ -0,0 +1,137 @@
...
+    class RequestWorker {
+        +const Available() bool
+        +AsyncDeliver(EventBatch, delivery_callback)
+    }
+
+    class WorkerPool {
+        +Get(worker_callback) void
+    }
+
+    class Summarizer {
```
kinyoklion, line 75: It seems like, in the implementation, the summarizer is also the data representation for the summary event? Maybe needs some means to indicate that, because I initially thought we were just missing summary events after looking at the diagram.

architecture/event_processor.md
```diff
@@ -0,0 +1,137 @@
...
+
+The Event Processor is responsible for consuming, batching, and delivering events generated
+by the server and client-side LaunchDarkly SDKs.
+
+```mermaid
+classDiagram
+    IEventProcessor <|-- NullEventProcessor
+    IEventProcessor <|-- AsioEventProcessor
+
+    AsioEventProcessor *-- LRUCache
```
kinyoklion, line 11: I think, in my diagrams, I had aggregate arrows the wrong direction. I think in your diagram you have associations the wrong direction.

image
From: https://www.omg.org/spec/UML/2.5.1/PDF

cwaldren-ld: I really wish the mermaid docs went into it with a quick reference, but they don't. And I see why they don't, it's a huge task.

architecture/event_processor.md
cwaldren-ld: That's a good point. Perhaps I should just add a `SummaryEvent` to make this intuitive, it wouldn't really change anything - let me try it.

cwaldren-ld: The change didn't really make sense. I'll document this instead in the diagram.

cwaldren-ld:
@kinyoklion Let me know if the summary events are clearer now.

PR #194, opened 2023-07-21
feat: Serialize flags and segments.

kinyoklion:


libs/internal/include/launchdarkly/data_model/flag.hpp
```diff
@@ -48,7 +49,7 @@ struct Flag {
         DEFINE_CONTEXT_KIND_FIELD(contextKind)
 
         Rollout() = default;
-        Rollout(std::vector<WeightedVariation>);
+        explicit Rollout(std::vector<WeightedVariation>);
```
kinyoklion, line 52: To prevent infinite template recursion.

libs/internal/src/serialization/json_flag.cpp
```diff
@@ -208,4 +208,160 @@ tag_invoke(boost::json::value_to_tag<
...
+
+    WriteMinimal(obj, "untracked", weighted_variation.untracked);
+}
+
+void tag_invoke(boost::json::value_from_tag const& unused,
+                boost::json::value& json_value,
+                data_model::Flag::Rollout::Kind const& kind) {
+    switch (kind) {
+        case Flag::Rollout::Kind::kUnrecognized:
+            // TODO: Should we be preserving the original string.
```
kinyoklion, line 228: I imagine all strongly typed SDKs have this problem. So I will check some others. I am not super happy with it. (Though generally I don't like having to serialize, but it doesn't seem like there is any real alternative.)

.github/workflows/client.yml
```diff
@@ -30,14 +30,14 @@ jobs:
           # Inform the test harness of test service's port.
           test_service_port: ${{ env.TEST_SERVICE_PORT }}
           extra_params: '-skip-from ./contract-tests/sdk-contract-tests/test-suppressions.txt'
-  build-test:
```
kinyoklion, line 33: I changed the build names to it is easier to scan them in the PR checks.

libs/internal/src/serialization/json_flag.cpp
```diff
@@ -208,4 +208,160 @@ tag_invoke(boost::json::value_to_tag<
...
+    auto& obj = json_value.emplace_object();
+
+    obj.emplace("variations", boost::json::value_from(rollout.variations));
+    if (rollout.kind != Flag::Rollout::Kind::kUnrecognized) {
+        // TODO: Should we be preserving the original string and putting it in.
+        obj.emplace("kind", boost::json::value_from(rollout.kind));
+    }
+    WriteMinimal(obj, "seed", rollout.seed);
+    if (rollout.bucketBy.Valid()) {
+        obj.emplace("bucketBy", rollout.bucketBy.RedactionName());
```
cwaldren-ld, line 251: q: Should be preserving this even if invalid?

kinyoklion: Maybe?

kinyoklion: I am fine with preserving it, because here we will have some logical value retained. Probably.

cwaldren-ld (approved):

PR #197, opened 2023-08-16
feat: server-side contract tests

cwaldren-ld:
Introduces a contract test server for the server-side SDK, runs in CI, and initial (large) suppression list.

The bulk of the code is copied from the client-side, although I have made an effort to refactor the common JSON definitions into a shared library. 

There is an opportunity to refactor out the "shell" (the coroutine stuff that is responsible for serving/responding to requests) into a common lib as well, but it didn't seem worth the effort.


.github/workflows/server.yml
```diff
@@ -6,11 +6,30 @@ on:
     paths-ignore:
       - '**.md' #Do not need to run CI for markdown changes.
   pull_request:
-    branches: [ main, server-side ]
+    branches: [ main, server-side, cw/sc-206686/server-build-ci ]
```
cwaldren-ld, line 9: TODO: remove. Needed to run CI since I rebased this PR from `server-side` to the `server-build-ci` branch.

libs/common/include/launchdarkly/data/evaluation_detail.hpp
```diff
@@ -60,6 +60,14 @@ class EvaluationDetail {
...
     [[nodiscard]] std::optional<EvaluationReason> const& Reason() const;
 
+    /**
+     * Check if an evaluation reason exists, and if so, if it is of a particular
+     * kind.
+     * @param kind Kind to check.
+     * @return True if a reason exists and matches the given kind.
+     */
+    [[nodiscard]] bool ReasonKindIs(enum EvaluationReason::Kind kind) const;
+
```
cwaldren-ld, line 70: I found myself needing to reach into the reason and check if it was a specific kind, like:
```c++
if (detail.Reason() && detail.Reason()->Kind() == EvaluationReason::Kind::kWhatever)
```
I feel like this could be a reasonably common task and deserves a method, but perhaps not.

libs/internal/include/launchdarkly/data_model/flag.hpp
```diff
@@ -88,7 +88,7 @@ struct Flag {
     std::vector<Target> targets;
     std::vector<Target> contextTargets;
     std::vector<Rule> rules;
-    std::optional<std::uint64_t> offVariation;
+    std::optional<std::int64_t> offVariation;
```
cwaldren-ld, line 91: Our JSON specification notes that the minimum values are 0, so I used `uint64_t` - but that's not the case in practice, as the test harness may send negative values.

libs/server-sdk/src/data_sources/data_source_event_handler.cpp
```diff
@@ -62,7 +62,7 @@ tl::expected<std::optional<DataSourceEventHandler::Put>, JsonError> tag_invoke(
     auto const& obj = json_value.as_object();
     PARSE_FIELD(path, obj, "path");
     // We don't know what to do with a path other than "/".
-    if (path != "/") {
+    if (!(path == "/" || path.empty())) {
```
cwaldren-ld, line 65: The path may be omitted in PUT events.

libs/server-sdk/src/evaluation/evaluator.cpp
```diff
@@ -138,7 +138,7 @@ EvaluationDetail<Value> Evaluator::FlagVariation(
     Flag const& flag,
     Flag::Variation variation_index,
     EvaluationReason reason) const {
-    if (variation_index >= flag.variations.size()) {
+    if (variation_index < 0 || variation_index >= flag.variations.size()) {
```
cwaldren-ld, line 141: Now that variation_index is signed, must check for negative.

libs/server-sdk/src/client_impl.cpp
```diff
@@ -200,7 +203,21 @@ template <typename T>
 EvaluationDetail<T> ClientImpl::VariationInternal(Context const& ctx,
```
cwaldren-ld, line 203: All the grossness in this method that follows is fixed in a future PR (https://github.com/launchdarkly/cpp-sdks/pull/203)

libs/common/include/launchdarkly/data/evaluation_detail.hpp
kinyoklion: I suppose it doesn't hurt anything.

libs/server-sdk/src/client_impl.cpp
```diff
@@ -240,10 +268,53 @@ EvaluationDetail<T> ClientImpl::VariationInternal(Context const& ctx,
...
+        event_processor_->SendAsync(std::move(event));
         return EvaluationDetail<T>(std::move(default_value), std::nullopt,
                                    error_reason);
     }
 
+    event.value = detail.Value();
+    event.variation = detail.VariationIndex();
+    event.version = flag.Version();
+
+    if (detailed) {
```
kinyoklion, line 283: I think we need another layer here to put the event together. Typically we have an event factory. Then you either make two factories and choose between them based on detail, or you could just have a parameter to a method.

 As is it seems like too much detail at this level.

Also, I think this needs additional experiment handling, but possibly that comes in a later PR.

kinyoklion (approved):

libs/server-sdk/src/client_impl.cpp
cwaldren-ld: (fixed in future PR)

PR #201, opened 2023-08-18
feat: implement AllFlagsState

cwaldren-ld:
Implements the `AllFlagsState` API, which evaluates all flags and returns a serializable structure suitable for bootstrapping a client-side SDK. 

The method returns an `AllFlagsState` class, which has a nested `State` class representing details for a single flag. Additionally, there is an `Options` enum to control the API's behavior.


libs/internal/include/launchdarkly/data_model/flag.hpp
```diff
@@ -16,8 +16,10 @@
 namespace launchdarkly::data_model {
 
 struct Flag {
-    using Variation = std::uint64_t;
-    using Weight = std::uint64_t;
+    using Variation = std::int64_t;
+    using Weight = std::int64_t;
+    using FlagVersion = std::uint64_t;
```
cwaldren-ld, line 21: Can't fall this `Version` because it conflicts with the `Version()` method. Hmm.

libs/server-sdk/src/client_impl.cpp
```diff
@@ -157,15 +158,50 @@ bool ClientImpl::Initialized() const {
...
-    return result;
+
+    if (config_.Offline()) {
+        LD_LOG(logger_, LogLevel::kWarn)
+            << "AllFlagsState() called, but client is in offline mode. "
+               "Returning empty state";
+        return {};
+    }
+
+    if (!Initialized()) {
```
kinyoklion, line 172: This flow doesn't look right taking persistence into account. 

If the flags are in persistence, then we would evaluate all flags and provide a warning that it was called before evaluation. 

If persistence wasn't populated, and we were not initialized, then I would expect this. The variation internal method has it.

Node equivalent
```
    if (!this.initialized()) {
      const storeInitialized = await this.featureStore.initialized();
      if (storeInitialized) {
        this.logger?.warn(
          'Called allFlagsState before client initialization; using last known' +
            ' values from data store',
        );
      } else {
        this.logger?.warn(
          'Called allFlagsState before client initialization. Data store not available; ' +
            'returning empty state',
        );
        valid = false;
      }
    }
```

kinyoklion: Variation internal has something similar.

kinyoklion (changes requested):

kinyoklion (approved):

PR #203, opened 2023-08-23
fix: refactor Variation methods for correctness & clarity

cwaldren-ld:
This significantly refactors the `Variation` internal calls to be much easier to reason about. In the process, it fixes some broken behavior which is now reflected in the suppressions list. 

It also refactors event generation to use the `EventScope`/`EventFactory` pattern found in other SDKs, such as Rust and Go.


libs/internal/src/data_model/flag.cpp
```diff
@@ -26,4 +26,26 @@ Flag::Rollout::Rollout(std::vector<WeightedVariation> variations_)
       bucketBy("key"),
       contextKind("user") {}
 
+bool Flag::IsExperimentationEnabled(
```
cwaldren-ld, line 29: Just moving this from where the `AllFlagsState` needed it internally into a public place, since it's used more broadly now.

libs/server-sdk/src/client_impl.cpp
```diff
@@ -125,8 +127,12 @@ static bool IsInitialized(DataSourceStatus::DataSourceState state) {
...
 void ClientImpl::Identify(Context context) {
-    event_processor_->SendAsync(events::IdentifyEventParams{
-        std::chrono::system_clock::now(), std::move(context)});
+    if (events_default_.disabled) {
+        return;
+    }
+
+    event_processor_->SendAsync(
+        events_default_.factory.Identify(std::move(context)));
 }
```
cwaldren-ld, line 136: There's a slight bit of weirdness here. It'd be nice to not have to do _any_ work if events are disabled - like is done here with the early out.

This skips any copies or constructors that would otherwise be called and then subsequently discarded by the `NullEventProcessor`, if events are disabled.

But then - why not just always construct a real event processor?
- Don't waste a thread servicing the real event processor's timer
- Speeds up instantiation of the SDK probably by a tiny amount
- Ensures events are not sent, event if `EventScope.disabled` isn't checked

Could also remove this boolean altogether and just keep relying on the `NullEventProcessor`. I copied this technique from how we do it in Rust.

libs/server-sdk/src/client_impl.cpp
```diff
@@ -234,193 +240,239 @@ void ClientImpl::FlushAsync() {
     event_processor_->FlushAsync();
 }
 
-template <typename T>
```
cwaldren-ld, line 237: This has been totally refactored to more closely match our specification. There's now a Pre and Post step, which take care of all the details including sending events.

libs/server-sdk/src/events/event_scope.hpp
```diff
@@ -0,0 +1,24 @@
+#pragma once
+
+#include <launchdarkly/events/event_processor_interface.hpp>
+
+#include "event_factory.hpp"
+
+namespace launchdarkly::server_side {
+
+class EventScope {
```
cwaldren-ld, line 9: This is slightly different than the Rust implementation because we rely on the `NullEventProcessor` pattern in this SDK. 

There's no need for a bool to check if the processor is enabled or not, before sending the event. 

But, there is an advantage to the Rust way of doing things - no need to construct the event arguments only to throw them away. 

I'm not sure this is a real problem though, perhaps the compiler can determine that we never reassign the runtime EventProcessor object and just optimize the whole thing away.

contract-tests/server-contract-tests/test-suppressions.txt
```diff
@@ -1,111 +1,3 @@
-# SC-214431
```
cwaldren-ld, line 1: Woohoo! Just streaming issues left.

kinyoklion (approved):

PR #204, opened 2023-08-23
feat: implement prerequisite event recording

cwaldren-ld:
Implements prerequisite event recording, and fixes some bugs related to prereq event generation.

### DIFF
--- a/libs/server-sdk/src/client_impl.cpp
+++ b/libs/server-sdk/src/client_impl.cpp
@@ -122,8 +122,8 @@ ClientImpl::ClientImpl(Config config, std::string const& version)
                                           http_properties_,
                                           logger_)),
       evaluator_(logger_, memory_store_),
-      events_default_(*event_processor_.get(), EventFactory::WithoutReasons()),
-      events_with_reasons_(*event_processor_.get(),
+      events_default_(event_processor_.get(), EventFactory::WithoutReasons()),
+      events_with_reasons_(event_processor_.get(),
                            EventFactory::WithReasons()) {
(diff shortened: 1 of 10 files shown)


libs/server-sdk/src/client_impl.cpp
```diff
@@ -80,7 +79,7 @@ bool EventsEnabled(Config const& config) {
     return config.Events().Enabled() && !config.Offline();
 }
```
cwaldren-ld, line 81: Decided to remove `NullEventProcessor` entirely from the server-side SDK, and just return a `nullptr` instead. We only need to check it in two places:
- `EventScope::Send`
- `Client::FlushAsync`

libs/server-sdk/src/evaluation/evaluator.cpp
```diff
@@ -65,16 +67,21 @@ EvaluationDetail<Value> Evaluator::Evaluate(
...
                                 EvaluationReason::PrerequisiteFailed(p.key));
             }
         }
     } else {
-        LogError(parent_key, Error::CyclicPrerequisiteReference(flag.key));
-        return OffValue(flag, EvaluationReason::MalformedFlag());
+        LogError(parent_key.value_or("(no parent)"),
+                 Error::CyclicPrerequisiteReference(flag.key));
+        return EvaluationReason::MalformedFlag();
     }
```
cwaldren-ld, line 85: This was a bug - it was returning the off variation, but should have instead returned a nullopt.

kinyoklion (approved):

PR #211, opened 2023-08-25
feat: move client-side config headers to client-sdk

cwaldren-ld:
This is the first in a series of PRs to re-organize the config related C bindings. 

Why now? 

We need the client and server SDKs to have different configs, because different options apply. At the same time, some config is sharable - such as loggers. The original 1.0 release of the client-side SDK mixed these concerns unintentionally. 

This PR focuses on moving the client-side `LDClientConfig` and `LDClientConfigBuilder` out of `common` and into `client-sdk`. 

This should be backwards compatible enough for a minor release, with some callouts in the release notes.  Although the header locations are moving, no symbol name or function/struct definition is changing. 

For `config.h`:
- Because this is included by `sdk.h`, we can change the path. Upgrade pain should be zero, unless a user has included it elsewhere in their application (without `sdk.h`.)

For `builder.h`:
- Because `sdk.h` _doesn't_ include the `config_builder.h`, people would have had to manually include it. Those builds will now break because the header has moved. 

To ease this transition, I've kept stub headers in the original locations which log messages using `#pragma message`. They can't automatically include the new location headers because that'd introduce a circular dependency. 

--------------------------------
Future work includes potentially renaming the sub-builders (such as `LDDataSourceStreamBuilder`) to include the `Client` prefix (e.g. `LDClientDataSourceStreamBuilder`). 

We could achieve this with some typedefs + forwarding functions in a backwards compatible manner.


libs/common/include/launchdarkly/bindings/c/config/builder.h
```diff
@@ -1,597 +1,16 @@
...
-/**
- * Frees the builder; only necessary if not calling Build.
- * @param builder Builder to free.
- */
-LD_EXPORT(void)
-LDClientConfigBuilder_Free(LDClientConfigBuilder builder);
-
-#ifdef __cplusplus
-}
-#endif
```
cwaldren-ld, line 595: So, this file still exists but just includes the new one with a compiler warning. I'm not sure we actually need to do this to remain backwards compatible in the traditional sense, but it is somewhat nicer than a build just breaking.

Why is this an issue at all? Because the main `sdk.h` binding file doesn't include the config builder bindings, since none of the types in that file need it. Users need to include it themselves. Could revisit that decision.

libs/common/include/launchdarkly/bindings/c/config/client_side/builder.h
```diff
@@ -0,0 +1,486 @@
+/** @file */
```
cwaldren-ld, line 1: This is almost the same file as before, except all the Custom/Basic logging builders have been extracted out into logging_builder.h. The intention is to be able to include that in the server-side config builder.

examples/hello-c-client/main.c
```diff
@@ -1,7 +1,7 @@
 #include <launchdarkly/client_side/bindings/c/sdk.h>
 
-#include <launchdarkly/bindings/c/config/builder.h>
 #include <launchdarkly/bindings/c/context_builder.h>
+#include <launchdarkly/client_side/bindings/c/config/builder.h>
```
cwaldren-ld, line 5: This demonstrates the customer change necessary: append `client_side` to the left of `/bindings`.

libs/client-sdk/include/launchdarkly/client_side/bindings/c/config/builder.h
```diff
@@ -0,0 +1,487 @@
+/** @file */
```
cwaldren-ld, line 1: This is mostly moving the old `builder.h` to this new location. Log related config was removed and remains in common.

libs/client-sdk/include/launchdarkly/client_side/bindings/c/config/config.h
```diff
@@ -0,0 +1,27 @@
+/** @file */
```
cwaldren-ld, line 1: This is just moving the file from the old location.

libs/client-sdk/src/CMakeLists.txt
```diff
@@ -43,21 +45,21 @@ add_library(${LIBNAME}
         flag_manager/flag_persistence.cpp
         bindings/c/sdk.cpp)
 
-if(MSVC OR (NOT BUILD_SHARED_LIBS))
```
cwaldren-ld, line 46: Looks like this got auto formatted.

libs/common/include/launchdarkly/bindings/c/config/builder.h
```diff
@@ -1,597 +1,14 @@
...
+ * The original C++ Client-side 1.0 release shipped builder.h at this location,
+ * but it was moved to the client_side subdirectory to accommodate the C++
+ * Server-side SDK. */
+#pragma once
 
-// NOLINTEND modernize-use-using
```
cwaldren-ld, line 597: Here are the messages.

libs/common/include/launchdarkly/bindings/c/config/config.h
```diff
@@ -1,27 +1,14 @@
...
- * be freed.
- * @param config Config to free.
- */
-LD_EXPORT(void) LDClientConfig_Free(LDClientConfig config);
-
-#ifdef __cplusplus
-}
-#endif
-
-// NOLINTEND modernize-use-using
```
cwaldren-ld, line 27: More messages.

libs/common/include/launchdarkly/bindings/c/config/logging_builder.h
```diff
@@ -0,0 +1,134 @@
+/** @file */
+// NOLINTBEGIN modernize-use-using
```
cwaldren-ld, line 2: This was extracted out of `sdk.h`. It can remain in common because it'll be shared by the client and server.

kinyoklion (approved):

PR #212, opened 2023-08-25
fix: adding slsa provenance generation to release-client workflow

rsoberano-ld:
Using [generic SLSA provenance generator ](https://github.com/slsa-framework/slsa-github-generator/blob/main/internal/builders/generic/README.md#a-different-attestation-for-each-iteration) to build provenance attestations for our client SDK artifacts. [Tech spec for reference ](https://launchdarkly.atlassian.net/wiki/spaces/SEC/pages/2529919025/Integrating+SLSA+compliance+with+SDK+release+workflows#Provenance-only-generator)

Since we're using a matrix strategy for the build, there will be one attestation file for all the artifacts for each platform (i.e., attestation for Linux artifacts, one for Windows artifacts, one for MacOS artifacts). 

Based off of POC here: https://github.com/launchdarkly/learn-release-please/pull/35/files though this time it should be more accurate since we're hashing the exact artifacts that we're uploading to the Github release, whereas there wasn't a good way to do that with the yarn-generated package in the POC repo.

### DIFF
--- a/.github/actions/sdk-release/action.yml
+++ b/.github/actions/sdk-release/action.yml
@@ -18,6 +18,16 @@ inputs:
     required: true
   sdk_cmake_target:
     description: 'CMake target of the sdk, e.g. launchdarkly-cpp-client.'
+outputs:
+  hashes-linux:
+    description: "base64-encoded sha256 hash of linux build artifacts"
+    value: ${{ steps.hash-linux.outputs.hashes-linux }}
+  hashes-windows:
+    description: "base64-encoded sha256 hash of windows build artifacts"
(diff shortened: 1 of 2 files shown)


rsoberano-ld:
Is there a good way to test this build workflow out for the cpp-sdks repo? I'm not 100% confident that `sha256sum ZIPFILES > base64 -w0` will work the same on all runner OSes, and it'd also be able to see this run all the way through to ensure it generates provenance properly.

cwaldren-ld:
Hey @rsoberano-ld , answered in PMs.

cwaldren-ld (approved):

PR #217, opened 2023-08-28
feat: add DataSourceStatus() and plumb through IClient interface

cwaldren-ld:
This was commented out / not present in the API before, but the implementation is now in the client so we just need expose it.


cwaldren-ld:
Merging as there is no logic change here.

PR #218, opened 2023-08-28
fix: initialization of LDFlagListener 

cwaldren-ld:
This commit moves the definition and implementation of `LDFlagListener` into `common`, so it can be easily shared with the server.

Additionally, it fixes a bug in the specification of the `LDFlagListener_Init` method. It previously took `LDFlagListener` by value; now it takes a pointer. 

The impact would be if a customer forgot to initialize the struct, then the member(s) would be undefined.

### DIFF
--- a/libs/common/include/launchdarkly/bindings/c/flag_listener.h
+++ b/libs/common/include/launchdarkly/bindings/c/flag_listener.h
@@ -4,6 +4,7 @@
 #pragma once
 
 #include <launchdarkly/bindings/c/export.h>
+#include <launchdarkly/bindings/c/value.h>
 
 #ifdef __cplusplus
 extern "C" {  // only need to export C interface if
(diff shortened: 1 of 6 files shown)


libs/common/include/launchdarkly/bindings/c/flag_listener.h
```diff
@@ -0,0 +1,64 @@
...
+ * Initializes a flag listener. Must be called before passing the listener
+ * to LDClientSDK_FlagNotifier_OnFlagChange.
+ *
+ * Create the struct, initialize the struct, set the FlagChanged handler
+ * and optionally UserData, and then pass the struct to
+ * LDClientSDK_FlagNotifier_OnFlagChange.
+ *
+ * @param listener Listener to initialize.
+ */
+LD_EXPORT(void) LDFlagListener_Init(struct LDFlagListener* listener);
```
cwaldren-ld, line 58: Signature has changed from `(struct LDFlagListener listener)`

kinyoklion (approved):
I think it is fine, it was a bug.

PR #224, opened 2023-08-29
fix: deserializing certain values leads to infinite loop

cwaldren-ld:
We had an infinite loop where the `Value` tag invoke deserializer was calling the `tl::expected<Value, JsonError` version, which called back to itself.

This was manifesting as a runtime infinite loop. 

It's not caught by contract tests since I believe the trigger is having a list of variations that are themselves arrays, like:
```
"variations": [["a","b","c",123,true,"false"], [ .. other stuff ..], ... ]
```
The fix is to break the cycle and remove the incorrect implementation.

### DIFF
--- a/libs/internal/src/serialization/json_value.cpp
+++ b/libs/internal/src/serialization/json_value.cpp
@@ -104,11 +104,5 @@ void tag_invoke(boost::json::value_from_tag const&,
     }
 }
 
-tl::expected<Value, JsonError> tag_invoke(
-    boost::json::value_to_tag<tl::expected<Value, JsonError>> const& tag,
-    boost::json::value const& json_value) {
-    return boost::json::value_to<Value>(json_value);
-}
-
 // NOLINTEND modernize-return-braced-init-list
 }  // namespace launchdarkly


libs/common/tests/value_test.cpp
```diff
@@ -219,28 +220,38 @@ TEST(ValueTests, OstreamOperator) {
         ProduceString(Value::Object{{"first", "cheese"}, {"second", 42}}));
 }
 
+Value ValueOrNull(boost::json::value value) {
```
cwaldren-ld, line 223: This is purely a convenience for these unit tests.

libs/internal/src/serialization/json_value.cpp
```diff
@@ -104,11 +104,5 @@ void tag_invoke(boost::json::value_from_tag const&,
     }
 }
 
-tl::expected<Value, JsonError> tag_invoke(
```
cwaldren-ld, line 107: This wasn't exposed in the json_value header, so I think I introduced it accidentally in a previous change.

cwaldren-ld:
I'm going to merge this one since it's causing all my upstream server-side-based PRs to be un-testable, and the change is very small.

PR #228, opened 2023-08-31
fix: ensure flag updates go through data store updater

cwaldren-ld:
I hadn't wired up the `DataStoreUpdater` component, so we were just inserting flag updates directly in the memory store. 

With this in place, we can remove the streaming update suppression.


libs/server-sdk/src/client_impl.cpp
```diff
@@ -111,10 +111,11 @@ ClientImpl::ClientImpl(Config config, std::string const& version)
       work_(boost::asio::make_work_guard(ioc_)),
       memory_store_(),
       status_manager_(),
```
cwaldren-ld, line 113: This is saying: store updates in the `memory_store`, and also read flag data from the `memory_store`.

kinyoklion (approved):

PR #229, opened 2023-08-31
fix: allow for specification of initial reconnect delay in streaming data source

cwaldren-ld:
The public API was already available, but we didn't have the plumbing required to get it into the SSE client. Therefore, I'm regarding it as a bugfix. 

This came up because we need it to pass some streaming contract tests, which in part fail because our default delays are too high (starting at 1 second).


kinyoklion (commented):
I feel like there are some todos associated with this.

kinyoklion:
https://github.com/launchdarkly/cpp-sdks/blob/054df6f02266e3c54e414cf5a83f251d1c0922e0/libs/client-sdk/src/data_sources/streaming_data_source.cpp#L94

kinyoklion:
I think some code should be needed in the streaming data source?

kinyoklion (approved):

PR #234, opened 2023-09-05
chore: resolve ValidChar function lint/compile warnings

cwaldren-ld:
The argument to `ValidChar` is a `char`, so its max value is `255` rendering the check redundant. This causes a warning on our Mac compilations as well as a linter warning. 

(unless we have a platform that has a char wider than 8 bits..)

### DIFF
--- a/libs/common/src/config/app_info_builder.cpp
+++ b/libs/common/src/config/app_info_builder.cpp
@@ -22,7 +22,7 @@ tl::expected<std::string, Error> AppInfoBuilder::Tag::Build() const {
 
 bool ValidChar(char c) {
     if (c > 0) {
-        // The MSVC implementation of isalnum will assert if the number it
+        // The MSVC implementation of isalnum will assert if the number is
         // outside its lookup table (0-0xFF, inclusive.)
         // iswalnum would not, but is less restrictive than desired.
         return std::isalnum(c) != 0 || c == '-' || c == '.' || c == '_';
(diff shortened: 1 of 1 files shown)


libs/common/src/config/app_info_builder.cpp
```diff
@@ -21,9 +21,9 @@ tl::expected<std::string, Error> AppInfoBuilder::Tag::Build() const {
 }
 
 bool ValidChar(char c) {
-    if(c > 0 && c < 255) {
```
cwaldren-ld, line 24: Could put a `static_assert(std::numeric_limits<char>::max() <= 0xFF)`. Is that too paranoid?

libs/common/src/config/app_info_builder.cpp
```diff
@@ -21,9 +21,9 @@ tl::expected<std::string, Error> AppInfoBuilder::Tag::Build() const {
...
-        // The MSVC implementation of isalnum will assert if the number it outside
-        // its lookup table.
+    if (c > 0) {
+        // The MSVC implementation of isalnum will assert if the number it
```
kinyoklion, line 25: I must have typoed this 'it'. Should be 'is'.

kinyoklion, line 25:
```suggestion
        // The MSVC implementation of isalnum will assert if the number is
```

kinyoklion (approved):

PR #237, opened 2023-09-07
fix: instances of undefined behavior in enum matches

cwaldren-ld:
We're currently using the pattern of "assert if enum match isn't exhaustive", but that's potentially invoking undefined behavior when the assertions are compiled out (if our code is wrong.)

This adds return values for some switches triggering the linter warnings.

### DIFF
--- a/libs/common/src/bindings/c/value.cpp
+++ b/libs/common/src/bindings/c/value.cpp
@@ -4,6 +4,7 @@
 #include <launchdarkly/bindings/c/value.h>
 #include <launchdarkly/bindings/c/iter.hpp>
 #include <launchdarkly/detail/c_binding_helpers.hpp>
+#include <launchdarkly/detail/unreachable.hpp>
 #include <launchdarkly/value.hpp>
 
 using launchdarkly::Value;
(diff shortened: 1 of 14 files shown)


libs/common/src/bindings/c/value.cpp
```diff
@@ -58,8 +58,10 @@ LD_EXPORT(enum LDValueType) LDValue_Type(LDValue val) {
             return LDValueType_Object;
         case Value::Type::kArray:
             return LDValueType_Array;
+        default:
```
kinyoklion, line 61: I feel this has a pretty steep impact on API usage. Is it any better than the null type?

cwaldren-ld: Yea, after reflection this is not optimal. We have an underlying `variant` which already has a type tag; let me try taking advantage of that instead with an exhaustive visitor. Maybe the static analysis will be satisfied.

cwaldren-ld:
Going to leave this in a limbo state while I investigate better ways to enforce exhaustive matching. One technique used in the latest commit is to use __buildin_unreachable and `-werror=switch`.

kinyoklion (approved):

Issue #243, opened 2023-09-12
Client stops receiving flag updates after about 5 minutes


jdegges:
**Describe the bug**
  After about 5 minutes the launch darkly client will stop receiving flag updates through any registered listeners using `client.FlagNotifier().OnFlagChange()` as well as through a variation query like `client.BoolVariation(FEATURE_FLAG_KEY, false)`.
  
  **To reproduce**
* Checkout the repository, commit 6876f69e1d47f9d39e6a2110b7af66f69e32923f
* Modify the hello-c-client or hello-cpp-client examples to continuously query flag values by applying a diff like:
```
@@ -59,11 +60,14 @@ int main() {
         return 1;
     }
 
+    while (true) {
     bool flag_value =
         LDClientSDK_BoolVariation(client, FEATURE_FLAG_KEY, false);
 
     printf("*** Feature flag '%s' is %s for this user\n\n", FEATURE_FLAG_KEY,
            flag_value ? "true" : "false");
+    sleep(60);
+    }
```
* Build and run as usual: `mkdir build && cd build && cmake .. && make -j && ./examples/hello-c-client/hello-c` (or `./examples/hello-cpp-client/hello-cpp`)
* Toggle the flag value via the web ui about once every minute and observe the value change on the console
* After about 5 minutes the console will stop reflecting the most current flag value
  
**Expected behavior**
* The launch darkly client is capable of remaining connected and receiving the latest flag values.
* When the launch darkly client becomes disconnected or is otherwise unable to obtain latest flag values it notifies the caller of the situation so it can know what is happening and potentially take action.

**Logs**
Can provide logs upon request
  
**SDK version**
Commit 6876f69e1d47f9d39e6a2110b7af66f69e32923f
  
**Language version, developer tools**
Apple clang version 12.0.0 (clang-1200.0.32.29)
gcc v8
  
**OS/platform**
MacOS 13.4
Linux
  
**Additional context**
This behavior was reproduced using both the C++ and C interfaces of this cpp-sdks project on both MacOS and Linux using both the variation query interface, the all flags interface, and the OnFlagChange listener interface. In all attempts the client stopped working after about 5 minutes.

When the examples were instrumented with a `OnDataSourceStatusChange()` listener the listener callback was never executed.

When periodic polling of `client.Initialized()` was added to the examples it always returned "1".

[...]

PR #244, opened 2023-09-12
fix: stream connections longer than 5 minutes are dropped 

cwaldren-ld:
This fixes [streaming connection dropping after 5 minutes](https://github.com/launchdarkly/cpp-sdks/issues/243). 

The problem was applying a 5 minute timeout to an `async_read` _operation_, when the actual intent was to apply it to the act of receiving _any data_. 

The timeout exists to prevent the client from hanging infinitely if the server stops responding. LaunchDarkly sends heartbeats to clear the timer.

With `async_read`, the operation won't complete until the response body is finished. Since the body will never finish until the client shuts down (or server interrupts it), it's not possible to institute a timeout. 

The solution is to use `async_read_some`, which completes whenever data arrives.

This way, we can assign the 5 minute timeout as an upper bound on receiving any data, as was originally intended. 

-----

A side effect of this change was breaking the `reconnection` SSE contract tests, which proved to be somewhat of a rabbithole. 

The upshot is that we weren't handling chunked encoding properly when chunked encoding _ends_. The original code assumes we either terminate an async read with an error, or it goes on forever. 

In fact it should perform the backoff algorithm if we detect the end of chunked encoding.

-----

I've also included more comments, more debug logging, and renamed some symbols for clarity. This should aid in debugging issues like this more efficiently. 

- [ ] Long running connection test (4 hours no drops)

### DIFF
--- a/contract-tests/sse-contract-tests/include/event_outbox.hpp
+++ b/contract-tests/sse-contract-tests/include/event_outbox.hpp
@@ -72,11 +72,11 @@ class EventOutbox : public std::enable_shared_from_this<EventOutbox> {
    private:
     RequestType build_request(
         std::size_t counter,
-        std::variant<launchdarkly::sse::Event, launchdarkly::sse::Error> ev);
+        std::variant<launchdarkly::sse::Event, launchdarkly::sse::Error> event);
     void on_resolve(beast::error_code ec, tcp::resolver::results_type results);
     void on_connect(beast::error_code ec,
                     tcp::resolver::results_type::endpoint_type);
     void on_flush_timer(boost::system::error_code ec);
     void on_write(beast::error_code ec, std::size_t);
-    void do_shutdown(beast::error_code ec, std::string what);
+    void do_shutdown(beast::error_code ec);
 };
--- a/contract-tests/sse-contract-tests/src/entity_manager.cpp
+++ b/contract-tests/sse-contract-tests/src/entity_manager.cpp
@@ -27,7 +27,7 @@ std::optional<std::string> EntityManager::create(ConfigParams const& params) {
     }
 
     if (params.body) {
-        client_builder.body(std::move(*params.body));
+        client_builder.body(*params.body);
     }
 
     if (params.readTimeoutMs) {
@@ -44,12 +44,13 @@ std::optional<std::string> EntityManager::create(ConfigParams const& params) {
         LD_LOG(logger_, LogLevel::kDebug) << std::move(msg);
     });
 
-    client_builder.receiver([copy = poster](launchdarkly::sse::Event e) {
-        copy->post_event(std::move(e));
+    client_builder.receiver([copy = poster](launchdarkly::sse::Event event) {
+        copy->post_event(std::move(event));
     });
 
-    client_builder.errors(
-        [copy = poster](launchdarkly::sse::Error e) { copy->post_error(e); });
+    client_builder.errors([copy = poster](launchdarkly::sse::Error event) {
+        copy->post_error(event);
+    });
 
     auto client = client_builder.build();
     if (!client) {
--- a/contract-tests/sse-contract-tests/src/entity_manager.cpp
+++ b/contract-tests/sse-contract-tests/src/entity_manager.cpp
@@ -35,6 +35,11 @@ std::optional<std::string> EntityManager::create(ConfigParams const& params) {
             std::chrono::milliseconds(*params.readTimeoutMs));
     }
 
+    if (params.initialDelayMs) {
+        client_builder.initial_reconnect_delay(
+            std::chrono::milliseconds(*params.initialDelayMs));
+    }
+
     client_builder.logger([this](std::string msg) {
         LD_LOG(logger_, LogLevel::kDebug) << std::move(msg);
     });
--- a/contract-tests/sse-contract-tests/src/entity_manager.cpp
+++ b/contract-tests/sse-contract-tests/src/entity_manager.cpp
@@ -53,7 +53,7 @@ std::optional<std::string> EntityManager::create(ConfigParams const& params) {
         return std::nullopt;
     }
 
-    client->run();
+    client->async_connect();
 
     entities_.emplace(id, std::make_pair(client, poster));
     return id;
--- a/contract-tests/sse-contract-tests/src/event_outbox.cpp
+++ b/contract-tests/sse-contract-tests/src/event_outbox.cpp
@@ -93,6 +93,8 @@ EventOutbox::RequestType EventOutbox::build_request(
                         break;
                     case Error::UnrecoverableClientError:
                         msg.comment = "unrecoverable client error";
+                    case Error::ReadTimeout:
+                        msg.comment = "read timeout";
                     default:
                         msg.comment = "unspecified error";
                 }
--- a/contract-tests/sse-contract-tests/src/event_outbox.cpp
+++ b/contract-tests/sse-contract-tests/src/event_outbox.cpp
@@ -14,8 +14,6 @@ auto const kOutboxCapacity = 1023;
 EventOutbox::EventOutbox(net::any_io_executor executor,
                          std::string callback_url)
     : callback_url_{std::move(callback_url)},
-      callback_port_{},
-      callback_host_{},
       callback_counter_{0},
       executor_{executor},
       resolver_{executor},
@@ -29,7 +27,7 @@ EventOutbox::EventOutbox(net::any_io_executor executor,
     callback_port_ = uri_components->port();
 }
 
-void EventOutbox::do_shutdown(beast::error_code ec, std::string what) {
+void EventOutbox::do_shutdown(beast::error_code ec) {
     event_stream_.socket().shutdown(tcp::socket::shutdown_both, ec);
     flush_timer_.cancel();
 }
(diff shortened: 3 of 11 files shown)


contract-tests/sse-contract-tests/src/event_outbox.cpp
```diff
@@ -67,7 +67,7 @@ EventOutbox::RequestType EventOutbox::build_request(
     RequestType req;
 
     req.set(http::field::host, callback_host_);
-    req.method(http::verb::get);
+    req.method(http::verb::post);
     req.target(callback_url_ + "/" + std::to_string(counter));
```
cwaldren-ld, line 71: Noticed this during some wireshark spelunking and was very confused for a moment.

libs/client-sdk/src/data_sources/streaming_data_source.cpp
```diff
@@ -138,7 +138,7 @@ void StreamingDataSource::Start() {
 
     client_builder.logger([weak_self](auto msg) {
         if (auto self = weak_self.lock()) {
-            LD_LOG(self->logger_, LogLevel::kDebug) << msg;
+            LD_LOG(self->logger_, LogLevel::kDebug) << "sse-client: " << msg;
         }
```
cwaldren-ld, line 142: Quality of life improvement.

libs/common/src/config/logging_builder.cpp
```diff
@@ -60,7 +60,9 @@ LoggingBuilder::BasicLogging& LoggingBuilder::BasicLogging::Tag(
     return *this;
 }
 LoggingBuilder::BasicLogging::BasicLogging()
-    : level_(Defaults<AnySDK>::LogLevel()), tag_(Defaults<AnySDK>::LogTag()) {}
+    : level_(GetLogLevelEnum(std::getenv("LD_LOG_LEVEL"),
+                             Defaults<AnySDK>::LogLevel())),
+      tag_(Defaults<AnySDK>::LogTag()) {}
```
cwaldren-ld, line 65: Ideally, we can just run `LD_LOG_LEVEL=debug ./binary` but the Console Logger constructed in the client passes in a default value, so the environment variable isn't inspected.

Now it works like you'd want.

libs/server-sent-events/src/client.cpp
```diff
@@ -79,19 +79,45 @@ class FoxyClient : public Client,
...
+     * on the response's body has completed, and how long that operation took
+     * (using a monotonic clock.)
+     *
+     * This is useful for debugging timeout-related issues, like receiving a
+     * heartbeat message.
+     *
+     * The timestamp of the last read is initialized after receiving successful
+     * headers, but before initiating the first body read. Future reads then
+     * update the timestamp.
+     */
```
cwaldren-ld, line 108: Added this helper so we can more easily tell how fast / when data is being received. For example, this clearly shows heartbeats arriving every 180 seconds.

libs/server-sent-events/src/client.cpp
```diff
@@ -79,21 +79,49 @@ class FoxyClient : public Client,
           read_timeout_(read_timeout),
           write_timeout_(write_timeout),
           req_(std::move(req)),
```
cwaldren-ld, line 81: Re-arranging to match declaration order.

libs/server-sent-events/src/client.cpp
```diff
@@ -263,8 +293,22 @@ class FoxyClient : public Client,
     void on_read_body(boost::system::error_code ec, std::size_t amount) {
         boost::ignore_unused(amount);
         if (ec == boost::asio::error::operation_aborted) {
+            logger_("read HTTP response body aborted");
             return;
         }
```
cwaldren-ld, line 298: Here's the key: if the parser indicates that the chunked encoding is finished, initiate backoff instead of attempting to keep reading the body.

libs/server-sent-events/src/client.cpp
```diff
@@ -276,6 +320,7 @@ class FoxyClient : public Client,
     }
 
     void do_shutdown(std::function<void()> completion) {
```
cwaldren-ld, line 322: This was causing unnecessary hanging when the test service tells the client to close.

vendor/foxy/include/foxy/impl/session/async_read.impl.hpp
```diff
@@ -1,41 +1,66 @@
 //
```
cwaldren-ld, line 1: The majority of changes in these files is auto formatting, I only added an async_read_some wrapper.

Ideally this change would take place in our fork.

libs/server-sent-events/src/client.cpp
```diff
@@ -290,13 +299,13 @@ class FoxyClient : public Client,
 
     void on_read_body(boost::system::error_code ec, std::size_t amount) {
         boost::ignore_unused(amount);
-        if (ec == boost::asio::error::operation_aborted) {
-            logger_("read HTTP response body aborted");
+        if (ec == boost::asio::error::operation_aborted && shutting_down_) {
```
jdegges: If `on_read_body()` can ever be executed from a context that is different from where `do_shutdown()` is executed then we may have a race between modifying & accessing `shutting_down_`.

cwaldren-ld: This is a good callout, especially since I had to add the bool back. 

Although the public `async_shutdown` method will `post`'s to the `session_`'s executor, the `run` routine did not - I've fixed that.

Since those are the only two public entrypoints, we should be safe.

contract-tests/sse-contract-tests/src/entity_manager.cpp
```diff
@@ -35,6 +35,11 @@ std::optional<std::string> EntityManager::create(ConfigParams const& params) {
             std::chrono::milliseconds(*params.readTimeoutMs));
     }
```
cwaldren-ld, line 37: This was missing, but it's already implemented. Easy win for speeding up contract tests.

contract-tests/sse-contract-tests/include/event_outbox.hpp
```diff
@@ -72,11 +72,11 @@ class EventOutbox : public std::enable_shared_from_this<EventOutbox> {
    private:
```
cwaldren-ld, line 72: Lints.

contract-tests/sse-contract-tests/src/entity_manager.cpp
```diff
@@ -53,7 +59,7 @@ std::optional<std::string> EntityManager::create(ConfigParams const& params) {
         return std::nullopt;
     }
 
-    client->run();
```
cwaldren-ld, line 56: Renamed `run` to `async_connect` to make it clear this is kicking off an asynchronous operation.

contract-tests/sse-contract-tests/src/event_outbox.cpp
```diff
@@ -14,8 +14,6 @@ auto const kOutboxCapacity = 1023;
 EventOutbox::EventOutbox(net::any_io_executor executor,
                          std::string callback_url)
```
cwaldren-ld, line 15: Lints.

contract-tests/sse-contract-tests/src/event_outbox.cpp
```diff
@@ -54,20 +52,18 @@ void EventOutbox::run() {
...
 
 EventOutbox::RequestType EventOutbox::build_request(
     std::size_t counter,
-    std::variant<launchdarkly::sse::Event, launchdarkly::sse::Error> ev) {
+    std::variant<launchdarkly::sse::Event, launchdarkly::sse::Error> event) {
     RequestType req;
 
     req.set(http::field::host, callback_host_);
-    req.method(http::verb::get);
+    req.method(http::verb::post);
```
cwaldren-ld, line 66: `get` was just plain wrong, but accepted by the contract tests.

contract-tests/sse-contract-tests/src/main.cpp
```diff
@@ -38,6 +38,7 @@ int main(int argc, char* argv[]) {
         srv.add_capability("report");
         srv.add_capability("post");
         srv.add_capability("reconnection");
+        srv.add_capability("read-timeout");
```
cwaldren-ld, line 41: We were already passing the parameter along, just hadn't advertised the capability.

libs/server-sent-events/src/client.cpp
```diff
@@ -79,41 +79,48 @@ class FoxyClient : public Client,
...
               initial_reconnect_delay.value_or(kDefaultInitialReconnectDelay),
               kDefaultMaxBackoffDelay),
-          last_event_id_(std::nullopt),
-          backoff_timer_(session_.get_executor()),
-          event_receiver_(std::move(receiver)),
-          logger_(std::move(logger)),
-          errors_(std::move(errors)) {
+          backoff_timer_(std::move(executor)),
+          last_read_(std::nullopt),
+          shutting_down_(false) {
```
cwaldren-ld, line 93: Session construction now takes place in `create_session()`, so the logic can be reused for each new connection.

libs/server-sent-events/src/client.cpp
```diff
@@ -141,11 +149,17 @@ class FoxyClient : public Client,
         if (ec == boost::asio::error::operation_aborted) {
             return;
         }
-        run();
+        do_run();
     }
 
-    void run() override {
-        session_.async_connect(
```
cwaldren-ld, line 148: After renaming `run`, I wrapped it in a `post` so we can ensure there is only one thread accessing `session`. This is the same as `async_shutdown`.

vendor/foxy/include/foxy/impl/session/async_read.impl.hpp
kinyoklion: I think it is important not to auto format any of this code. It will make reconciling it with anything very annoying.

kinyoklion (changes requested):

vendor/foxy/include/foxy/impl/session/async_read.impl.hpp
cwaldren-ld: Great point. Fixed.

kinyoklion (approved):

PR #251, opened 2023-09-20
fix: catch exception if en_US.utf8-locale missing when parsing datetime headers

cwaldren-ld:
The event delivery subsystem inspects the HTTP headers on the event endpoints whenever it posts a batch of events.

It parses out the date header, storing it for later use. When emitting debug events, it is used to ensure that events stop emitting even if the host's time is off - as a form of time synchronization.

The parsing code is quite ugly, and also wrong - it assumed that `en_US.utf-8` would always be available. 

This commit hoists the locale loading routine out of the individual flush-workers and into the worker pool. If it can't load the locale, it emits a single `warn` log at startup explaining the effect.

### DIFF
--- a/libs/internal/include/launchdarkly/events/worker_pool.hpp
+++ b/libs/internal/include/launchdarkly/events/worker_pool.hpp
@@ -24,8 +24,6 @@ namespace launchdarkly::events {
  */
 class WorkerPool {
    public:
-    using ServerTimeCallback =
-        std::function<void(std::chrono::system_clock::time_point)>;
     /**
      * Constructs a new WorkerPool.
      * @param io The executor used for all workers.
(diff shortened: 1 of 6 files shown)


libs/internal/include/launchdarkly/events/worker_pool.hpp
```diff
@@ -24,8 +24,6 @@ namespace launchdarkly::events {
  */
 class WorkerPool {
    public:
-    using ServerTimeCallback =
-        std::function<void(std::chrono::system_clock::time_point)>;
     /**
```
cwaldren-ld, line 27: (unused)

kinyoklion (approved):

cwaldren-ld:
Well this is odd. With no changes to CI, apparently the windows build can't find OpenSSL. Looking into it.

PR #253, opened 2023-09-22
fix: treat warnings as errors in CI

cwaldren-ld:
Many users might compile the SDK using a "warnings as errors" flag. 

It would be helpful to enable that flag in our CI, such that we can be relatively sure that the SDK compiles w/o warnings on major compilers.

### DIFF
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -17,6 +17,6 @@ cd build
 # script ends.
 trap cleanup EXIT
 
-cmake -G Ninja -D BUILD_TESTING="$2" ..
+cmake -G Ninja -DCMAKE_COMPILE_WARNINGS_AS_ERROR=TRUE -D BUILD_TESTING="$2" ..
 
 cmake --build . --target "$1"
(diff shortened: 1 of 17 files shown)


cwaldren-ld:
Alright, looks like we're now compiling with warnings-as-errors in CI. To be clear, this is CI only - would be very annoying to have this be part of the normal development experience.

scripts/build.sh
```diff
@@ -17,6 +17,6 @@ cd build
 # script ends.
 trap cleanup EXIT
 
-cmake -G Ninja -D BUILD_TESTING="$2" ..
+cmake -G Ninja -DCMAKE_COMPILE_WARNING_AS_ERROR=TRUE -D BUILD_TESTING="$2" ..
```
cwaldren-ld, line 20: [Reference](https://cmake.org/cmake/help/latest/prop_tgt/COMPILE_WARNING_AS_ERROR.html)

kinyoklion (approved):

PR #255, opened 2023-10-09
build: clean up LD CMake variables & allow for OpenSSL dynamic link

cwaldren-ld:


cwaldren-ld:
Note: right now the shared library flag doesn't work due to foxy (in theory.) Needs some more work.

CMakeLists.txt
```diff
@@ -13,22 +14,63 @@ project(
...
+# to link against that statically or dynamically.
+
+option(LD_STATIC_LINK_OPENSSL
+        "Link OpenSSL as a static library or as a shared library"
+        ON  # default to linking OpenSSL statically
+)
+
+option(LD_BUILD_EXAMPLES "Build hello-world examples." ON)
+
+
```
cwaldren-ld, line 56: Throwing a warning when someone tries to build shared lib with unit tests enabled. This won't work since we hide symbols by default.

CMakeLists.txt
```diff
@@ -53,13 +95,22 @@ if (BUILD_TESTING)
...
 
+# Even though the main SDK might be a static or shared lib, boost should always statically
+# linked into the binary.
 set(Boost_USE_STATIC_LIBS ON)
 
-if (BUILD_SHARED_LIBS)
+if (NOT LD_BUILD_STATIC_LIBS)
     # When building a shared library we hide all symbols
     # aside from this we have specifically exported for the C-API.
     set(CMAKE_CXX_VISIBILITY_PRESET hidden)
```
kinyoklion, line 116: Maybe not right now, but we probably need an option for this. For those that build themselves and want to build a C++ dll.

kinyoklion: (If they build all their own stuff, then the ABI issue isn't really an SDK issue.)

libs/client-sdk/src/CMakeLists.txt
```diff
@@ -43,14 +49,16 @@ add_library(${LIBNAME}
...
         flag_manager/flag_manager.cpp
         flag_manager/flag_persistence.cpp
-        bindings/c/sdk.cpp)
+        bindings/c/sdk.cpp
+)
+
 
-if (MSVC OR (NOT BUILD_SHARED_LIBS))
+if (MSVC OR LD_BUILD_STATIC_LIBS)
     target_link_libraries(${LIBNAME}
```
kinyoklion, line 57: We may want to note this MSVC condition in the readme.

Unless we want to figure out how to make it actually work with dynamically linked boost on windows (world of pain).

kinyoklion (approved):

PR #256, opened 2023-10-09
chore: detect where OpenSSL is installed on Windows runner

cwaldren-ld:
We have intermittent CI failures because OpenSSL is installed to either `Program Files\OpenSSL-Win64` or `Program Files\OpenSSL`. I'm not entirely sure why, but it seems to be luck of the draw on the runner.

This should detect where it was installed and pass it to CMake. 

We can't force the install location because that requires a licensed chocolatey. Chocolatey also writes out the install location - for this particular package - but we'd have to process its output. Instead, this just tests the two known installation directories.

- [x] Tested with manual release

### DIFF
--- a/.github/actions/sdk-release/action.yml
+++ b/.github/actions/sdk-release/action.yml
@@ -103,10 +103,8 @@ runs:
       run: |
         if [ -d "C:\Program Files\OpenSSL-Win64" ]; then
           echo "OPENSSL_ROOT_DIR=C:\Program Files\OpenSSL-Win64" >> "$GITHUB_ENV"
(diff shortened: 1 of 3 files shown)


.github/actions/sdk-release/action.yml
```diff
@@ -97,12 +97,20 @@ runs:
       shell: bash
       run: |
         choco upgrade openssl --no-progress
-
+    - name: Determine OpenSSL Installation Directory
```
kinyoklion, line 100: This is kind of depressing, but I get it.

kinyoklion (approved):

PR #260, opened 2023-10-18
fix: LD_BUILD_SHARED_LIBS build flag usage

cwaldren-ld:
In my previous PR cleaning up cmake variables, during development I renamed the `BUILD_SHARED_LIBS` flag to `BUILD_STATIC_LIBS`, but then reconsidered and stuck with the original naming.

Unfortunately I didn't un-rename some usages of it - so we ended up having some instances of `LD_BUILD_STATIC_LIBS` in our `CMakeLists.txt`.

It appears this didn't have an obvious impact on our build/release process because:
1) Since `NOT LD_BUILD_STATIC_LIBS` would always be true (variable not defined), then it would have been setting `-fvisibility=hidden` on both shared and static builds. I'm pretty sure that flag has no affect when building static libs so it was a no-op in that situation.
2) Using the `target_sources` trick for pulling in the boost libraries always works whether we're building static libs or dynamic libs, so another no-op. 

It'd be nice to test this going forward. Some ideas:
- Generate a report on the artifact sizes, like the C Server SDK 
- Generate a report corresponding to something like `nm -gU artifact.dylib | wc -l` to get the number of symbols (e.g. 8k when visibility=hidden, 26k without)
- Force the SDK to be compiled as a shared lib with unit testing enabled, and assert that the compilation fails due to missing symbols 

Manual workflow run:
- [x] https://github.com/launchdarkly/cpp-sdks/actions/runs/6565100278

### DIFF
--- a/libs/client-sdk/src/CMakeLists.txt
+++ b/libs/client-sdk/src/CMakeLists.txt
@@ -76,7 +76,7 @@ set_property(TARGET ${LIBNAME} PROPERTY
         MSVC_RUNTIME_LIBRARY "MultiThreaded$<$<CONFIG:Debug>:Debug>DLL")
 
 install(TARGETS ${LIBNAME})
-if (BUILD_SHARED_LIBS AND MSVC)
+if (LD_BUILD_SHARED_LIBS AND MSVC)
     install(FILES $<TARGET_PDB_FILE:${LIBNAME}> DESTINATION bin OPTIONAL)
 endif ()
 # Using PUBLIC_HEADERS would flatten the include.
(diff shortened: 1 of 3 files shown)


libs/client-sdk/src/CMakeLists.txt
```diff
@@ -76,7 +76,7 @@ set_property(TARGET ${LIBNAME} PROPERTY
         MSVC_RUNTIME_LIBRARY "MultiThreaded$<$<CONFIG:Debug>:Debug>DLL")
 
 install(TARGETS ${LIBNAME})
-if (BUILD_SHARED_LIBS AND MSVC)
+if (LD_BUILD_SHARED_LIBS AND MSVC)
     install(FILES $<TARGET_PDB_FILE:${LIBNAME}> DESTINATION bin OPTIONAL)
```
cwaldren-ld, line 80: We don't have `cmake` tests that actually try to use the install files, which is why this didn't break CI.

louis-launchdarkly (approved):

PR #263, opened 2023-10-19
fix: enforce validated ContextKind usage in Segment data model

cwaldren-ld:
Resolves an outstanding TODO in `server-side` by enforcing segments to deserialize with valid context kinds.

### DIFF
--- a/libs/internal/src/serialization/json_segment.cpp
+++ b/libs/internal/src/serialization/json_segment.cpp
@@ -20,7 +20,10 @@ tl::expected<std::optional<data_model::Segment::Target>, JsonError> tag_invoke(
 
     data_model::Segment::Target target{};
 
-    PARSE_FIELD_DEFAULT(target.contextKind, obj, "contextKind", "user");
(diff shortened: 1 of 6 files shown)


libs/internal/src/serialization/json_segment.cpp
```diff
@@ -20,7 +20,10 @@ tl::expected<std::optional<data_model::Segment::Target>, JsonError> tag_invoke(
 
     data_model::Segment::Target target{};
 
-    PARSE_FIELD_DEFAULT(target.contextKind, obj, "contextKind", "user");
+    // The zero value of a ContextKind ("" - empty string) is not valid in the
+    // domain of possible contexts. This field is parsed as REQUIRED to
+    // specify that fact.
+    PARSE_REQUIRED_FIELD(target.contextKind, obj, "contextKind");
```
cwaldren-ld, line 26: This can be required since Targets will always have a contextKind. Users are added in the `included` or `excluded` fields (and omitted from `includedContexts` and `excludedContexts`).

kinyoklion (approved):

PR #265, opened 2023-10-20
chore: add data source status backwards-compat unit test

cwaldren-ld:
Adds a unit test that should catch if we make a breaking change to the types accessible via `client.DataSourceStatus()`.

This is useful because the upcoming server-side SDK refactors the Data Source Status types into the common library, and we must therefore provide aliases in the client SDK.


libs/client-sdk/tests/client_test.cpp
```diff
@@ -30,7 +30,7 @@ TEST(ClientTest, BoolVariationDefaultPassesThrough) {
     Client client(ConfigBuilder("sdk-123").Build().value(),
                   ContextBuilder().Kind("cat", "shadow").Build());
```
cwaldren-ld, line 32: Formatter took the reigns here.

kinyoklion (approved):

PR #268, opened 2023-10-23
chore: add server-side SDK to release-please config

cwaldren-ld:
Adds the server-side SDK into release please @ `0.1.0`. I've marked `prerelease: true`, so it should be marked as such when we do a Github release.

### DIFF
--- a/libs/server-sdk/package.json
+++ b/libs/server-sdk/package.json
@@ -4,7 +4,7 @@
   "version": "0.1.0",
   "private": true,
   "dependencies": {
-    "launchdarkly-cpp-internal": "0.1.6",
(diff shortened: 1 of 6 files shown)


libs/server-sdk/package.json
```diff
@@ -0,0 +1,10 @@
+{
+  "name": "launchdarkly-cpp-server",
+  "description": "This package.json exists for modeling dependencies for the release process.",
+  "version": "0.1.0",
+  "private": true,
+  "dependencies": {
```
cwaldren-ld, line 6: TODO: bump to latest

kinyoklion (approved):

PR #272, opened 2023-10-23
chore: remove server-side version from release-please manifest

cwaldren-ld:
This is an attempt to fix the pending release. 

Ideally, release please would detect that no release is necessary or that a 0.1.0 initial release of the server side only is necessary.

Right now, it thinks that the `sse-client` needs a release, and it's therefore bumping that and the server/client SDKs.

### DIFF
--- a/.release-please-manifest.json
+++ b/.release-please-manifest.json
@@ -1,6 +1,5 @@
 {
   "libs/client-sdk": "3.2.1",
-  "libs/server-sdk": "0.1.0",
   "libs/server-sent-events": "0.2.0",
   "libs/common": "0.5.0",
   "libs/internal": "0.3.0"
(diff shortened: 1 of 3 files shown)


.release-please-manifest.json
```diff
@@ -1,6 +1,5 @@
 {
   "libs/client-sdk": "3.2.1",
```
cwaldren-ld, line 2: Theory: release please will now:
(1) Not think that a release has been performed for `0.1.0` (which it hasn't) due to this change
(2) Will assign the next release number as `0.1.0` due to change in `release-please-config.json:23`

kinyoklion (approved):
This should make the server version you are wanting. 

I do not understand why all commits are being added for sse release though.

PR #280, opened 2023-10-27
chore: run client hello app in CI

cwaldren-ld:
So far we've been building the `hello-*` apps in CI, but not running them - meaning it's the responsibility of the SDK author.

This commit executes the hello app in CI, serving as an end-to-end test for:
- [ ] Static linking of the C bindings
- [ ] Static linking of native C++ 

Ideally we'd have such tests for multiple CMake configurations, such as linking as static vs dynamic library.

### DIFF
--- a/.github/actions/ci/action.yml
+++ b/.github/actions/ci/action.yml
@@ -15,10 +15,16 @@ inputs:
     required: false
     default: 'true'
   platform_version:
+    description: 'Boost platform version'
     required: false
     default: "22.04"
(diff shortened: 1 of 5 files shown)


cwaldren-ld:
Thinking about this, I'd rather have a true matrix build {dynamic link, static link} x {windows, mac, linux}.

That way we can build only 3 times, do it in parallel, and execute the hello tests serially for on each runner.

PR #282, opened 2023-10-27
chore: refactor openssl setup into dedicated x-platform action

cwaldren-ld:
Setting up the correct OpenSSL version should be encapsulated in the `ci` action, but it leaked out when I added the code to ensure OpenSSL > 3 on all runners.

Putting it back in its place. Motivated because I'm making another action that could use this.

### DIFF
--- a/.github/actions/install-openssl/action.yml
+++ b/.github/actions/install-openssl/action.yml
@@ -1,0 +1,32 @@
+name: Install OpenSSL
+description: 'Install OpenSSL >= 3 if not already present.'
+
+runs:
+  using: composite
+  steps:
+    # The ubuntu runner already has OpenSSL > 3 and CMake can find it.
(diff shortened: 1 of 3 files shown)


.github/actions/install-openssl/action.yml
```diff
@@ -0,0 +1,32 @@
+name: Install OpenSSL
+description: 'Install OpenSSL >= 3 if not already present.'
+
+runs:
+  using: composite
+  steps:
+    # The ubuntu runner already has OpenSSL > 3 and CMake can find it.
+    - name: Install for Ubuntu.
```
cwaldren-ld, line 8: The OpenSSL version on ubuntu 22.04 isn't going to go backwards (unless entropy in the universe reverses), so I didn't do anything smart here.

kinyoklion (approved):

PR #283, opened 2023-11-03
refactor!: move server side config into lib/server

cwaldren-ld:
Our current config system utilizes templated builders for different components, but then has a single top-level `Config` shared between client and server. This served us well during development of the server alpha, but it's now time to split the configs. 

This change moves the serve config into `lib/server`, in preparation for adding new configuration builders/structs.


libs/common/tests/config_builder_test.cpp
```diff
@@ -20,14 +19,6 @@ TEST_F(ConfigBuilderTest, DefaultConstruction_ClientConfig) {
     ASSERT_EQ(cfg->SdkKey(), "sdk-123");
 }
```
cwaldren-ld, line 21: Moved into `lib/server`'s tests.

cwaldren-ld:
Not entirely sure what `feat!` will do here, but my hypothesis is that it'll result in:
- Minor bump for the server (since we have `bump-minor-pre-major`)
- Major bump for common lib
- ..triggering a patch to client?

contract-tests/server-contract-tests/src/entity_manager.cpp
```diff
@@ -1,9 +1,9 @@
 #include "entity_manager.hpp"
 #include <boost/json/parse.hpp>
```
kinyoklion, line 2: Linter is angry in this file.

kinyoklion (approved):

PR #285, opened 2023-11-03
chore: fix OpenSSL root dir

cwaldren-ld:
We seem to be getting Boost 1.82 or 1.83 luck of the draw on Mac runners. Unfortunately, 1.83 has a breaking change - Boost.url is no longer header-only. 

The existing script was wrong anyways, since it was assuming `brew install boost` would get 1.82.

### DIFF
--- a/.github/actions/install-boost/action.yml
+++ b/.github/actions/install-boost/action.yml
@@ -39,8 +39,8 @@ runs:
       if: runner.os == 'macOS'
       shell: bash
       run: |
-        brew install boost@1.83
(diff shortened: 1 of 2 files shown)


cwaldren-ld:
Ah, right. Brew only keeps the latest version of boost 😢 .

cwaldren-ld:
Turns out the openssl action wasn't actually setting `OPENSSL_ROOT_DIR` this whole time - it just happened to work because CMake's find module was good enough.

PR #286, opened 2023-11-04
chore: set OpenSSL root dir properly in CI

cwaldren-ld:
In our "install OpenSSL" CI step, I hadn't properly setup the "return values" - setting the `OPENSSL_ROOT_DIR` properly. 

The builds still mostly worked, but were fragile. This is due to CMake's find modules finding them anyways. Except when they don't!

### DIFF
--- a/.github/actions/install-boost/action.yml
+++ b/.github/actions/install-boost/action.yml
@@ -39,8 +39,8 @@ runs:
       if: runner.os == 'macOS'
       shell: bash
       run: |
-        brew install boost@1.83
-        echo "BOOST_ROOT=$(brew --prefix boost@1.83)" >> $GITHUB_OUTPUT
+        brew install boost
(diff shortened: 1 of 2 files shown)


.github/actions/install-boost/action.yml
```diff
@@ -40,7 +40,7 @@ runs:
       shell: bash
       run: |
         brew install boost
-        echo "BOOST_ROOT=$(boost --prefix boost@1.82)" >> $GITHUB_OUTPUT
+        echo "BOOST_ROOT=$(brew --prefix boost)" >> $GITHUB_OUTPUT
```
cwaldren-ld, line 43: This was also just.. wrong. Luckily FindBoost appears to find it anyways.

kinyoklion (approved):

PR #289, opened 2023-11-16
draft for comparisons

cwaldren-ld:


libs/client-sdk/tests/data_source_status_manager_test.cpp
```diff
@@ -145,6 +147,8 @@ TEST(DataSourceStatusManagerTests, TimeIsUpdatedOnStateChange) {
     status_manager.SetState(DataSourceStatus::DataSourceState::kValid);
 
     auto initial = status_manager.Status().StateSince();
+
+    std::this_thread::sleep_for(std::chrono::milliseconds(1));
```
cwaldren-ld, line 151: TODO(add comment): This test randomly failed because the time hadn't changed at all. Not an ideal way to fix it, but we don't currently have a way to inject the time dependency.

libs/server-sdk/src/data_systems/lazy_load/lazy_load_system.hpp
```diff
@@ -72,16 +72,25 @@ class LazyLoad final : public data_interfaces::IDataSystem {
     void RefreshSegment(std::string const& key) const;
 
     template <typename TResult>
-    static TResult Get(data_components::ExpirationTracker::TrackState state,
-                       std::function<void(void)> const& refresh,
-                       std::function<TResult(void)> const& get) {
+    TResult Get(std::string const& key,
```
cwaldren-ld, line 75: The algorithm isn't as elegant now as it was with the fallthroughs, but I wanted to log the actual action taking place. Suppose I could use a string stream instead.

.github/workflows/server-redis.yml
```diff
@@ -0,0 +1,49 @@
...
+          cmake_target: launchdarkly-cpp-server-redis-source
+          simulate_release: true
+  build-redis-mac:
+    runs-on: macos-12
+    steps:
+      - uses: actions/checkout@v3
+      - uses: ./.github/actions/ci
+        with:
+          cmake_target: launchdarkly-cpp-server-redis-source
+          platform_version: 12
```
cwaldren-ld, line 34: Unless there's a quickfix for running Redis on the macos and windows runners, I'm going to file a ticket for adding it in the future.

libs/server-sdk-redis-source/src/redis_source.cpp
```diff
@@ -0,0 +1,75 @@
+#include <launchdarkly/server_side/integrations/redis/redis_source.hpp>
+
```
cwaldren-ld, line 2: Wondering if the clang diagnostic fails because it is somehow using cmake to build everything (and hence wouldn't know about adding `LD_BUILD_REDIS_SUPPORT`.)

scripts/build-release.sh
```diff
@@ -5,10 +5,18 @@
 
 set -e
```
cwaldren-ld, line 7: This is more a general issue that we build targets independently, but we require the full cmake configuration step to do so.

Might make sense to go further down this path and add `LD_BUILD_SERVER` and `LD_BUILD_CLIENT` so we can speed up CI / disable unecessary work for people who don't want it. Or, we could have different cmake projects and pull them together into a superbuild.

libs/server-sdk/src/data_systems/lazy_load/lazy_load_system.hpp
kinyoklion: A little sad, it was relatively pleasant looking for C++. :-P

cwaldren-ld: You're right, it was a jewel. Gonna try to fix it up with a string-stream.

libs/server-sdk/src/data_components/serialization_adapters/json_deserializer.hpp
```diff
@@ -1,31 +1,87 @@
...
+        return data_interfaces::IDataReader::StorageItem<Item>(
+            data_interfaces::IDataReader::Tombstone(descriptor.version));
+    }
+
+    auto const json_val = boost::json::parse(descriptor.serializedItem);
+
+    auto result =
+        boost::json::value_to<tl::expected<std::optional<Item>, JsonError>>(
+            json_val);
+
```
cwaldren-ld, line 27: TODO: make an actual tag_invoke deserializer for this.

libs/server-sdk/src/data_systems/lazy_load/lazy_load_system.hpp
cwaldren-ld: Elegance restored.

examples/hello-cpp-server-redis/main.cpp
```diff
@@ -0,0 +1,102 @@
...
+char const* get_with_env_fallback(char const* source_val,
+                                  char const* env_variable,
+                                  char const* error_msg);
+using namespace launchdarkly;
+using namespace launchdarkly::server_side;
+
+int main() {
+    char const* sdk_key = get_with_env_fallback(
+        SDK_KEY, "LD_SDK_KEY",
+        "Please edit main.c to set SDK_KEY to your LaunchDarkly server-side "
```
kinyoklion, line 29:
```suggestion
        "Please edit main.cpp to set SDK_KEY to your LaunchDarkly server-side "
```

examples/hello-cpp-server-redis/main.cpp
```diff
@@ -0,0 +1,102 @@
...
+        "SDK key "
+        "first.\n\nAlternatively, set the LD_SDK_KEY environment "
+        "variable.\n"
+        "The value of SDK_KEY in main.c takes priority over LD_SDK_KEY.");
+
+    auto config_builder = ConfigBuilder(sdk_key);
+
+    using LazyLoad = server_side::config::builders::LazyLoadBuilder;
+
+    auto redis = integrations::RedisDataSource::Create("redis://localhost:6379",
```
kinyoklion, line 39: Maybe a comment here instructing them to set this URL to something that will work for them.

kinyoklion (approved):

PR #290, opened 2023-11-16
feat: add data system interfaces [1]

cwaldren-ld:
This PR adds the initial set of interfaces defined in the Data System spec. The intention is to allow our internal components to interoperate using the same interfaces that a 3rd party would use to implement a custom persistent store.

Subsequent PRs will:
- Create BackgroundSync (streaming+polling) / LazyLoad (redis) data system implementations using these interfaces. 
- Add configuration for the data system
- Update the server to use these interfaces internally


libs/server-sdk/include/launchdarkly/server_side/integrations/iserialized_item_kind.hpp
```diff
@@ -0,0 +1,43 @@
+#pragma once
```
cwaldren-ld, line 1: This lives in the public `include` directory because it will need to be consumed by implementers of persistent data stores.

libs/server-sdk/include/launchdarkly/server_side/integrations/persistent_store_core.hpp
```diff
@@ -1,214 +0,0 @@
-#pragma once
```
cwaldren-ld, line 1: (refactored into two files)

cwaldren-ld:
It's expected for the mac build to fail at this time.

libs/server-sdk/include/launchdarkly/server_side/integrations/iserialized_item_kind.hpp
```diff
@@ -0,0 +1,44 @@
...
+ */
+class ISerializedItemKind {
+   public:
+    /**
+     * The namespace for the data.
+     */
+    [[nodiscard]] virtual std::string const& Namespace() const = 0;
+
+    /**
+     * Deserialize data and return the version of the data.
```
kinyoklion, line 21: I experienced a kind of deja vu, then realized I wrote these words.

libs/server-sdk/src/data_interfaces/destination/idestination.hpp
```diff
@@ -0,0 +1,52 @@
...
+
+/**
+ * \brief IDestination represents a sink for data received by the
+ * SDK. A destination may be a database, local file, etc.
+ */
+class IDestination {
+   public:
+    /**
+     * \brief Initialize the destination with a base set of data.
+     * \param data_set The initial data received by the SDK.
```
kinyoklion, line 18: Is there a reason for the syntax change? from `@` to `\`. I guess it is canonical.

libs/server-sdk/src/data_interfaces/source/ipush_source.hpp
```diff
@@ -0,0 +1,54 @@
...
+#include <launchdarkly/data_model/sdk_data_set.hpp>
+
+#include <functional>
+#include <optional>
+#include <string>
+
+namespace launchdarkly::server_side::data_interfaces {
+
+/**
+ * \brief IPushSource obtains data via a push synchronization mechanism,
```
kinyoklion, line 12: I think we need a more elaborate explanation on this one, because the name and the content don't make sense to me.

Maybe the actual pushing happens outside the interface, which is confusing.

libs/server-sdk/src/data_interfaces/system/isystem.hpp
```diff
@@ -0,0 +1,34 @@
...
+
+#include "../store/istore.hpp"
+
+namespace launchdarkly::server_side::data_interfaces {
+
+/**
+ * \brief ISystem obtains data used for flag evaluations and makes it available
+ * to other components.
+ */
+class ISystem : public IStore {
```
kinyoklion, line 11: Not sure about this name.

kinyoklion (approved):

libs/server-sdk/src/data_interfaces/destination/idestination.hpp
cwaldren-ld: CLion Nova is the reason. I failed to change the defaults. I will do so now, and can probably fix these all with a find/replace in the final feature PR.

libs/server-sdk/src/data_interfaces/system/isystem.hpp
cwaldren-ld: The `ISystem` name? I guess it should actually be `IDataSystem` (but you might have objections to that as well? It's from the Data System spec work.)

The intention behind this interface is to reflect that it might be more than an `IStore`. Perhaps that's overengineering, but it does add the `Initialize()` method, which `IStore()` doesn't have.

kinyoklion: It is like IManager. If you use the pattern more than once, then you have already name squatted. I know it has a namespace, but it would end up very confusing when actually reading code.

IDataSystem doesn't have that problem, at least not really for our SDKs.

libs/server-sdk/src/data_interfaces/destination/idestination.hpp
cwaldren-ld: Well, I can't find the normal option to change the Doxygen format. Maybe it's not implemented yet.

libs/server-sdk/src/data_interfaces/system/isystem.hpp
cwaldren-ld: Ah,  I originally had all of the interfaces with a `Data` prefix, like `IDataStore`, `IDataDestination`, etc. 

But when put in the `data_interface` namespace, these all seemed quite redundant. I've been using the namespace prefixes throughout in the callsites instead. `data_interfaces::IDataSystem` might be fine though since it is the top-level concept (and like you said, `System` is insanely generic.)

cwaldren-ld: Will rename to `IDataSystem` in subsequent PR.

PR #291, opened 2023-11-16
feat: refactor existing components into data system [2]

cwaldren-ld:
This PR moves the contents of `src/data_sources` (polling + streaming implementations) into dedicated subdirectories of a new directory, `data_components`. 

The thousand-yard view is that we now have a larger abstraction (Data System) of which these components are a part of.

The objects in `data_components` can be composed by SDK authors to create the Background Sync and Lazy Load data systems. Background Sync is the equivalent of streaming/polling, while Lazy Load operates against a database.

The next PR will refactor the existing polling and streaming data sources to use these (now relocated) components.


libs/server-sdk/src/data_components/dependency_tracker/tagged_data.hpp
```diff
@@ -0,0 +1,26 @@
+#pragma once
```
cwaldren-ld, line 1: I `git mv`'d this. Guess it wasn't enough.

kinyoklion (approved):

PR #292, opened 2023-11-17
feat: data system config builders [3]

cwaldren-ld:
This introduces a set of configuration objects & builders for the server-side Data System.

These are defined in `libs/server-sdk` rather than in common, since they don't apply to the client at the moment. In a future client major version, we could attempt to unify them.

### DIFF
--- a/libs/server-sdk/include/launchdarkly/server_side/config/builders/all_builders.hpp
+++ b/libs/server-sdk/include/launchdarkly/server_side/config/builders/all_builders.hpp
@@ -1,0 +1,27 @@
+#pragma once
+
+#include <launchdarkly/config/shared/builders/app_info_builder.hpp>
+#include <launchdarkly/config/shared/builders/endpoints_builder.hpp>
+#include <launchdarkly/config/shared/builders/events_builder.hpp>
+#include <launchdarkly/config/shared/builders/http_properties_builder.hpp>
+#include <launchdarkly/config/shared/builders/logging_builder.hpp>
+
+#include <launchdarkly/server_side/config/builders/data_system/background_sync_builder.hpp>
+#include <launchdarkly/server_side/config/builders/data_system/bootstrap_builder.hpp>
+#include <launchdarkly/server_side/config/builders/data_system/data_destination_builder.hpp>
+#include <launchdarkly/server_side/config/builders/data_system/data_system_builder.hpp>
+#include <launchdarkly/server_side/config/builders/data_system/lazy_load_builder.hpp>
+
+namespace launchdarkly::server_side::config::builders {
+
+using SDK = launchdarkly::config::shared::ServerSDK;
+using EndpointsBuilder =
+    launchdarkly::config::shared::builders::EndpointsBuilder<SDK>;
+using HttpPropertiesBuilder =
+    launchdarkly::config::shared::builders::HttpPropertiesBuilder<SDK>;
+using AppInfoBuilder = launchdarkly::config::shared::builders::AppInfoBuilder;
+using EventsBuilder =
+    launchdarkly::config::shared::builders::EventsBuilder<SDK>;
+using LoggingBuilder = launchdarkly::config::shared::builders::LoggingBuilder;
+
+}  // namespace launchdarkly::server_side::config::builders
--- a/libs/server-sdk/include/launchdarkly/server_side/config/builders/data_system/bootstrap_builder.hpp
+++ b/libs/server-sdk/include/launchdarkly/server_side/config/builders/data_system/bootstrap_builder.hpp
@@ -1,0 +1,18 @@
+#pragma once
+
+#include <launchdarkly/server_side/config/built/data_system/bootstrap_config.hpp>
+
+#include <optional>
+
+namespace launchdarkly::server_side::config::builders {
+
+class BootstrapBuilder {
+   public:
+    BootstrapBuilder();
+
+    [[nodiscard]] std::optional<built::BootstrapConfig> Build() const;
+
+   private:
+    std::optional<built::BootstrapConfig> config_;
+};
+}  // namespace launchdarkly::server_side::config::builders
--- a/libs/server-sdk/include/launchdarkly/server_side/config/builders/data_system/data_destination_builder.hpp
+++ b/libs/server-sdk/include/launchdarkly/server_side/config/builders/data_system/data_destination_builder.hpp
@@ -1,0 +1,16 @@
+#pragma once
+
+#include <launchdarkly/server_side/config/built/data_system/data_destination_config.hpp>
+
+namespace launchdarkly::server_side::config::builders {
+
+struct DataDestinationBuilder {
+    DataDestinationBuilder();
+
+    [[nodiscard]] built::DataDestinationConfig Build() const;
+
+   private:
+    built::DataDestinationConfig config_;
+};
+
+}  // namespace launchdarkly::server_side::config::builders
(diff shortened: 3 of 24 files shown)


libs/server-sdk/include/launchdarkly/server_side/config/builders/all_builders.hpp
```diff
@@ -0,0 +1,27 @@
+#pragma once
```
cwaldren-ld, line 1: This file pulls in all the shared builders, making the templated versions available within the `server_side::config::builders` namespace. At the same time it pulls in the locally defined builders, so they all exist at the same level.

libs/server-sdk/include/launchdarkly/server_side/config/builders/data_system/bootstrap_builder.hpp
```diff
@@ -0,0 +1,18 @@
+#pragma once
```
cwaldren-ld, line 1: This is mostly a placeholder and not implemented yet.

libs/server-sdk/include/launchdarkly/server_side/config/builders/data_system/data_destination_builder.hpp
```diff
@@ -0,0 +1,16 @@
+#pragma once
```
cwaldren-ld, line 1: Mostly a placeholder, not built yet.

libs/server-sdk/include/launchdarkly/server_side/config/builders/data_system/data_system_builder.hpp
```diff
@@ -0,0 +1,29 @@
...
+
+namespace launchdarkly::server_side::config::builders {
+
+class DataSystemBuilder {
+   public:
+    DataSystemBuilder();
+    using BackgroundSync = BackgroundSyncBuilder;
+    using LazyLoad = LazyLoadBuilder;
+
+    DataSystemBuilder& Disabled(bool disabled);
```
cwaldren-ld, line 17: Right now the status quo is to have an `offline` boolean that disables the data source + events. I was thinking it'd be nice to have a clearly labeled `Disabled()` call for just the sources:

```c++
auto cfg = ConfigBuilder("sdk123")
                      .DataSystem().Disabled(true)
```

That way we can say "Offline mode is shorthand for calling Events().Disabled() and DataSystem.Disabled()" or similar.

libs/server-sdk/include/launchdarkly/server_side/config/builders/data_system/lazy_load_builder.hpp
```diff
@@ -0,0 +1,65 @@
...
+     */
+    LazyLoadBuilder& CacheTTL(std::chrono::milliseconds ttl);
+
+    /**
+     * \brief Specify the eviction policy when a data item's TTL expires.
+     * At this time, only EvictionPolicy::Disabled is supported (the default),
+     * which leaves stale items in the cache until they can be refreshed. \param
+     * policy The EvictionPolicy. \return Reference to this.
+     */
+    LazyLoadBuilder& CacheEviction(EvictionPolicy policy);
```
cwaldren-ld, line 57: Ryan's existing work also had the Cache eviction interval and a mode for active eviction. We can have this, but I haven't added it here (though I suspect we could add it later just as well.)

libs/server-sdk/include/launchdarkly/server_side/config/built/data_system/bootstrap_config.hpp
```diff
@@ -0,0 +1,8 @@
+#pragma once
+
```
cwaldren-ld, line 2: Placeholder

libs/server-sdk/include/launchdarkly/server_side/config/built/data_system/data_destination_config.hpp
```diff
@@ -0,0 +1,9 @@
+#pragma once
```
cwaldren-ld, line 1: Placeholder

libs/server-sdk/include/launchdarkly/server_side/config/built/data_system/lazy_load_config.hpp
```diff
@@ -0,0 +1,25 @@
...
+     */
+    enum class EvictionPolicy {
+        /* No action taken; eviction is disabled. Stale items will be used
+         * in evaluations if they cannot be refreshed. */
+        Disabled = 0
+    };
+
+    EvictionPolicy eviction_policy;
+    std::chrono::milliseconds eviction_ttl;
+    std::shared_ptr<server_side::data_interfaces::ISerializedDataPullSource>
```
cwaldren-ld, line 22: This is the thing anyone could implement (`ISerializedDataPullSource`). 

The `shared_ptr` is pretty general - we can have a higher level constructor to make a Redis client object, or let people pass in their own.

We could also have `source` be a `variant<shared_ptr<ISerializedDataPullSource>, SomeRedisArgs>` instead if we wanted to defer construction to SDK creation time and totally encapsulate it. 

But, I think the current way is nice because then the Client itself has zero knowledge that it's dealing with Redis.

libs/server-sdk/include/launchdarkly/server_side/config/defaults.hpp
```diff
@@ -0,0 +1,10 @@
+#pragma once
```
cwaldren-ld, line 1: Placeholder.

libs/server-sdk/src/config/builders/data_system/data_system_builder.cpp
```diff
@@ -0,0 +1,50 @@
...
+    method_builder_ = std::move(lazy_load);
+    return *this;
+}
+
+DataSystemBuilder& DataSystemBuilder::Disabled(bool const disabled) {
+    config_.disabled = disabled;
+    return *this;
+}
+
+tl::expected<built::DataSystemConfig, Error> DataSystemBuilder::Build() const {
```
cwaldren-ld, line 24: This is pretty simple in concept but just gross to look at. 

The problem is unifying the return values of `LazyLoad::Build` (fallible, `tl::expected`) and `BackgroundSync::Build` (not fallible, since if no source is provided we default to streaming.)

Instead of doing it in build, could instead call `Build` in the `::Method` implementations. Then we'd need to store an:
```c++
std::variant<
  tl::expected<LazyLoadConfig, Error>, 
  tl::expected<BackgroundSyncConfig, Error>
>
```
Maybe not horrible 🤷

libs/server-sdk/src/config/config_builder.cpp
```diff
@@ -40,7 +34,7 @@ tl::expected<Config, Error> ConfigBuilder::Build() const {
     if (sdk_key.empty()) {
         return tl::make_unexpected(Error::kConfig_SDKKey_Empty);
     }
-    auto offline = offline_.value_or(Defaults::Offline());
```
cwaldren-ld, line 43: Ignore this. Had some back-and-forth thoughts about offline mode; will add it back in as a top-level config since it's convenient / status quo.

libs/server-sdk/src/config/config.cpp
```diff
@@ -2,54 +2,50 @@
 
 namespace launchdarkly::server_side {
 
+using namespace launchdarkly::config::shared;
+
 Config::Config(std::string sdk_key,
```
cwaldren-ld, line 7: Removing redundant namespace qualifiers + plumbing the new `DataSystemConfig`

keelerm84 (approved):

PR #294, opened 2023-11-17
refactor: existing data sources to use data system components [4]

cwaldren-ld:
This PR takes the existing components (like change notifier, expiration tracker, streaming/polling datasources) and refactors them to use the Data System interfaces.

The broad strokes of the architecture are now visible in the server's directory structure. 

There are a lot of details remaining to be filled in. Subsequent PRs!


libs/server-sdk/src/client_impl.cpp
```diff
@@ -74,57 +97,53 @@ static Logger MakeLogger(config::shared::built::Logging const& config) {
         std::make_shared<logging::ConsoleBackend>(config.level, config.tag)};
 }
```
cwaldren-ld, line 99: I removed the config.Offline, will re-add in subsequent PR.

libs/server-sdk/src/client_impl.cpp
```diff
@@ -177,24 +196,31 @@ AllFlagsState ClientImpl::AllFlagsState(Context const& context,
...
-                << "AllFlagsState() called before client has finished "
-                   "initializing; using last known values from data store";
-        } else {
-            LD_LOG(logger_, LogLevel::kWarn)
-                << "AllFlagsState() called before client has finished "
-                   "initializing. Data store not available. Returning empty "
-                   "state";
-            return {};
-        }
+        //        if (memory_store_.Initialized()) {
```
cwaldren-ld, line 199: This is commented out because I haven't yet implement the initialization states in the DS spec.

libs/server-sdk/src/client_impl.hpp
```diff
@@ -173,12 +166,12 @@ class ClientImpl : public IClient {
...
-
-    data_sources::DataSourceStatusManager status_manager_;
-    data_components::ChangeNotifier data_store_updater_;
+    data_components::DataSourceStatusManager status_manager_;
 
-    std::shared_ptr<::launchdarkly::data_sources::IDataSource> data_source_;
+    // This is the main polymorphic component that constitutes the
+    // guts of how data is retrieved (polling, streaming, persistent stores,
+    // etc.)
+    std::unique_ptr<data_interfaces::IDataSystem> data_system_;
```
cwaldren-ld, line 174: (renamed to `IDataSystem` as mentioned in previous PR)

libs/server-sdk/src/data_components/serialization_adapters/json_pull_source.hpp
```diff
@@ -0,0 +1,72 @@
...
+            return tl::make_unexpected(ErrorToString(flag.error()));
+        }
+
+        std::optional<DataModel> maybe_flag = flag->value();
+
+        if (!maybe_flag) {
+            /* JSON was valid, but the value is 'null'
+             * TODO: will this ever happen?
+             */
+            return tl::make_unexpected("data was null");
```
cwaldren-ld, line 61: This is the case where we get back a serialized item, but then upon deserializing the JSON, it's actually `null`.

The deserializer implementation (called from `value_to`) was designed to return an `std::optional` so the caller could distinguish "got a valid object" vs omitted where it is important to do so.

In this case, I don't think we'd ever store `null` in the persistent store so don't think it would come up in practice.

libs/server-sdk/src/data_interfaces/bootstrapper/ibootstrapper.hpp
```diff
@@ -0,0 +1,72 @@
...
+
+/**
+ * Defines a component that can fetch a complete dataset for use in a
+ * Data System. Bootstrapping takes place when the SDK starts, and is
+ * responsible for provisioning the initial data that an SDK uses before it can
+ * begin the on-going synchronization process.
+ */
+class IBootstrapper {
+   public:
+    class Error {
```
cwaldren-ld, line 22: This is a prototype of what the IBoostrapper interface might look like and isn't final. In practice I think we'd need `FetchAll` to be asynchronous (take a completion handler.)

libs/server-sdk/src/data_systems/lazy_load/lazy_load_system.cpp
```diff
@@ -0,0 +1,136 @@
...
+        }
+        if (data_components::ExpirationTracker::TrackState::kFresh == state) {
+            return initialized_.value();
+        }
+    }
+    RefreshInitState();
+    return initialized_.value_or(false);
+}
+
+void LazyLoad::RefreshAllFlags() const {
```
cwaldren-ld, line 81: These are all original TODOs from Ryan that I have not yet resolved.

libs/server-sdk/src/data_systems/lazy_load/lazy_load_system.cpp
```diff
@@ -0,0 +1,129 @@
...
+      time_([]() { return std::chrono::steady_clock::now(); }),
+      initialized_() {}
+
+std::string const& LazyLoad::Identity() const {
+    static std::string id = "lazy load via " + source_.Identity();
+    return id;
+}
+
+void LazyLoad::Initialize() {}
+
```
cwaldren-ld, line 26: Note: the logic here isn't complete, just copied over from what was already implemented (but not used.) 

This will need unit testing, coming in future PRs.

libs/server-sdk/src/data_interfaces/bootstrapper/ibootstrapper.hpp
```diff
@@ -0,0 +1,72 @@
...
+
+       private:
+        Error(Kind kind, std::optional<std::string> detail)
+            : kind(kind), detail(std::move(detail)) {}
+        Kind kind;
+        std::optional<std::string> detail;
+    };
+
+    /**
+     * Fetch a complete dataset. This method must invokable multiple times.
```
keelerm84, line 48: Can you clarify that second line?

libs/server-sdk/src/data_systems/background_sync/sources/README.md
```diff
@@ -0,0 +1,22 @@
...
+### Push Data Source
+
+The Push data source is able to operate in "streaming" or "polling" mode. In either mode, flag
+updates are delivered in the background from the perspective of the application.
+
+### Pull Data Source
+
+The Pull data source fetches flag data on-demand from a database. Because this is likely slow,
+it operates a read-through cache wherein each stored item has a TTL indicating when it should be
+re-fetched.
```
keelerm84, line 19: The terminology / distinctions here seem a little confusing.

Is the polling data model considered a "push data source"? That doesn't seem right. But the description of pull doesn't sound like you intended it to include polling.

keelerm84 (approved):

libs/server-sdk/src/data_systems/background_sync/sources/README.md
cwaldren-ld: Yea, discussed this with Ryan and I see how it's confusing. I'm trying to draw a distinction between the interface implemented by persistent stores, and that of our streaming/polling sources.

 
The terminology comes from observing the data flow from the source. With polling/streaming mode, the data is pushed into the memory cache - it's asynchronous. With the persistent stores, it's pulled into the memory cache on demand - synchronously. 

 I was looking for terms that:
- Are symmetric 
- Short
- Have a commonly known meaning
- Don't mention streaming/polling/persistent stores directly

libs/server-sdk/src/data_interfaces/bootstrapper/ibootstrapper.hpp
cwaldren-ld: Realized that info is extraneous, will remove in subsequent PR.

PR #295, opened 2023-11-21
refactor: rename IPush and IPull sources to better reflect their purposes

cwaldren-ld:
The original names were an attempting to retain some symmetry, while also showing that both modes are "sources".

This doesn't need to be the case. These "sources" are conceptual, but have totally different modes of interaction.

I've updated the names to `IDataSynchronizer` and `IDataReader`. These are names that I used in past revisions of the DS spec, and I think they are the better choice today.


libs/server-sdk/src/data_components/serialization_adapters/json_data_reader.cpp
```diff
@@ -0,0 +1,39 @@
+#include "json_data_reader.hpp"
```
cwaldren-ld, line 1: This is a file renaming, not new content (except for updating the `ISerializedDataReader` symbol.)

libs/server-sdk/src/data_components/serialization_adapters/json_data_reader.hpp
```diff
@@ -1,16 +1,15 @@
 #pragma once
```
cwaldren-ld, line 1: Since `IPullSource` is now an `IDataReader`, it makes sense that the `JsonSource` is now a `JsonDataReader`.

keelerm84 (approved):

PR #296, opened 2023-11-21
build: fix compilation errors in data system

cwaldren-ld:
This fixes a bunch of compiler errors on the `feat/data-system` branch, which existed because of the piecemeal manner in which I broke out chunks from my working Redis client branch.

### DIFF
--- a/libs/server-sdk/src/bindings/c/builder.cpp
+++ b/libs/server-sdk/src/bindings/c/builder.cpp
@@ -7,18 +7,19 @@
 #include <launchdarkly/server_side/config/config_builder.hpp>
 
 using namespace launchdarkly::server_side;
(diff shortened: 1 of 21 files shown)


libs/server-sdk/src/bindings/c/builder.cpp
```diff
@@ -121,7 +122,7 @@ LD_EXPORT(void)
 LDServerConfigBuilder_Offline(LDServerConfigBuilder b, bool offline) {
     LD_ASSERT_NOT_NULL(b);
```
cwaldren-ld, line 124: Will be fixed in subsequent PRs.

keelerm84 (approved):

PR #298, opened 2023-11-21
refactor: various client SDK lints

cwaldren-ld:
This is a mostly removing a bunch of redundant namespace qualifiers + formatting changes in the client SDK, which I addressed while building the Redis integration.

We can merge these directly to main instead to make the Redis PR cleaner.

### DIFF
--- a/libs/client-sdk/src/client_impl.cpp
+++ b/libs/client-sdk/src/client_impl.cpp
@@ -6,6 +6,7 @@
 #include <launchdarkly/events/asio_event_processor.hpp>
 #include <launchdarkly/events/null_event_processor.hpp>
 
+#include <launchdarkly/detail/c_binding_helpers.hpp>
 #include <launchdarkly/encoding/sha_256.hpp>
 #include <launchdarkly/logging/console_backend.hpp>
 #include <launchdarkly/logging/null_logger.hpp>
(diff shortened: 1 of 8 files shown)


libs/client-sdk/src/client_impl.cpp
```diff
@@ -30,14 +30,14 @@ using launchdarkly::client_side::data_sources::DataSourceStatus;
 using launchdarkly::config::shared::built::DataSourceConfig;
 using launchdarkly::config::shared::built::HttpProperties;
```
cwaldren-ld, line 32: Removing redundant namespace + formatting

libs/client-sdk/src/client_impl.hpp
```diff
@@ -1,33 +1,32 @@
 #pragma once
```
cwaldren-ld, line 2: Sorting include paths according to our standard from most specific to least specific.

libs/client-sdk/src/data_sources/data_source.hpp
```diff
@@ -0,0 +1,20 @@
+#pragma once
```
cwaldren-ld, line 1: Copying this from `internal` lib, as it will be removed when the server-side SDK is introduced and we don't want to port over the client immediately.

libs/client-sdk/src/data_sources/null_data_source.hpp
```diff
@@ -1,14 +1,13 @@
 #pragma once
```
cwaldren-ld, line 1: Sort includes + remove redundant namespace

libs/client-sdk/src/data_sources/polling_data_source.hpp
```diff
@@ -1,24 +1,24 @@
 #pragma once
```
cwaldren-ld, line 2: Ditto

libs/client-sdk/src/data_sources/streaming_data_source.hpp
```diff
@@ -1,10 +1,8 @@
 #pragma once
 
-#include <chrono>
```
cwaldren-ld, line 3: Remove useless `using` and sort includes + remove redundant namespaces

cwaldren-ld:
Mac CI is expected to fail due to a known issue with the updated boost libraries.

keelerm84 (approved):

PR #299, opened 2023-11-21
refactor: remove obsolete server-related common lib components

cwaldren-ld:
This removes obsolete server-related config from the `common` lib.


libs/common/tests/data_source_builder_test.cpp
```diff
@@ -1,6 +1,5 @@
 #include <gtest/gtest.h>
 #include <launchdarkly/config/client.hpp>
-#include <launchdarkly/config/server_builders.hpp>
 #include <launchdarkly/config/shared/sdks.hpp>
 #include <launchdarkly/value.hpp>
```
cwaldren-ld, line 5: (The removed tests are now in the server lib)

libs/common/tests/service_endpoints_test.cpp
```diff
@@ -1,7 +1,6 @@
 #include <gtest/gtest.h>
 
 #include <launchdarkly/config/client.hpp>
```
cwaldren-ld, line 3: (The removed tests are now in the server lib)

PR #303, opened 2023-11-22
feat: implement additional data system config builders

cwaldren-ld:
This adds in some missing top-level Data System builder methods, as well as additional documentation. 

Some config tests that I previously removed from the `common` lib are added back into the server. 

I've also re-added Offline mode configuration, which modifies Data System and Events configuration.

### DIFF
--- a/libs/server-sdk/src/config/config_builder.cpp
+++ b/libs/server-sdk/src/config/config_builder.cpp
@@ -3,7 +3,7 @@
 namespace launchdarkly::server_side {
 
 ConfigBuilder::ConfigBuilder(std::string sdk_key)
-    : sdk_key_(std::move(sdk_key)) {}
+    : sdk_key_(std::move(sdk_key)), offline_(false) {}
 
 config::builders::EndpointsBuilder& ConfigBuilder::ServiceEndpoints() {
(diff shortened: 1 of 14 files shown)


libs/server-sdk/src/config/config_builder.cpp
```diff
@@ -39,14 +44,23 @@ tl::expected<Config, Error> ConfigBuilder::Build() const {
     if (!endpoints_config) {
         return tl::make_unexpected(endpoints_config.error());
     }
```
cwaldren-ld, line 46: In this scheme, users can set `Offline` as a top-level config property, which is the same as disabling events and the data system. If they need more granularity, then can individually enable/disable just events or the data system.

libs/server-sdk/tests/service_endpoint_builders_test.cpp
```diff
@@ -0,0 +1,88 @@
+#include <gtest/gtest.h>
+
```
cwaldren-ld, line 2: (These were previously in the `common` lib, now moved into the server.)

keelerm84 (approved):

PR #306, opened 2023-11-22
refactor: improve IDataSynchronizer's interface to expose IDestination

cwaldren-ld:
This refactors `IDataSynchronizer` to take an `IDestination`  and bootstrap data in its `StartAsync` method, rather than having an independent `Init` function.

The addition of `IDestination` makes the interface's purpose clearer.

Before the destination was hidden - you'd only see it when constructing an implementation. Now it's clear that the `IDataSynchronizer` puts data into the `IDestination`; no hidden dependency.

It also doesn't make sense to "Init" an `IDataSynchronizer` with an SDKDataSet. 

The purpose of that call is to allow the synchronizer to inspect any existing data in possession by the SDK. The result of that inspection might mean an optimized fetch of initial data is possible (e.g. if there was an etag present in the initial data.)

### DIFF
--- a/libs/server-sdk/src/data_systems/background_sync/background_sync_system.cpp
+++ b/libs/server-sdk/src/data_systems/background_sync/background_sync_system.cpp
@@ -21,15 +21,15 @@ BackgroundSync::BackgroundSync(
                                          config::built::BackgroundSyncConfig::
                                              StreamingConfig>) {
                 synchronizer_ = std::make_shared<StreamingDataSource>(
-                    endpoints, method_config, http_properties, ioc,
-                    change_notifier_, status_manager, logger);
+                    ioc, logger, status_manager, endpoints, method_config,
+                    http_properties);
 
             } else if constexpr (std::is_same_v<
                                      T, config::built::BackgroundSyncConfig::
                                             PollingConfig>) {
                 synchronizer_ = std::make_shared<PollingDataSource>(
-                    endpoints, method_config, http_properties, ioc,
-                    change_notifier_, status_manager, logger);
+                    ioc, logger, status_manager, endpoints, method_config,
+                    http_properties);
             }
         },
         background_sync_config.synchronizer_);
@@ -43,9 +43,8 @@ BackgroundSync::BackgroundSync(
       synchronizer_(std::make_shared<NullDataSource>(ioc, status_manager)) {}
 
 void BackgroundSync::Initialize() {
-    // TODO: if there was any data from bootstrapping, then add it:
-    // synchronizer_->Init(data);
-    synchronizer_->StartAsync();
+    synchronizer_->StartAsync(&change_notifier_,
+                              nullptr /* no bootstrap data supported yet */);
 }
 
 std::string const& BackgroundSync::Identity() const {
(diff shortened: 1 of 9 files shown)


libs/server-sdk/src/data_systems/background_sync/background_sync_system.cpp
```diff
@@ -21,15 +21,15 @@ BackgroundSync::BackgroundSync(
                                          config::built::BackgroundSyncConfig::
                                              StreamingConfig>) {
```
cwaldren-ld, line 22: Re-arranged the constructor arguments to follow our C++ standards.

libs/server-sdk/src/data_systems/background_sync/sources/polling/polling_data_source.cpp
```diff
@@ -44,22 +43,20 @@ std::string const& PollingDataSource::Identity() const {
 }
 
 PollingDataSource::PollingDataSource(
+    boost::asio::any_io_executor const& ioc,
```
cwaldren-ld, line 46: These changes follow from re-arranging the constructor args.

libs/server-sdk/src/data_systems/background_sync/sources/streaming/streaming_data_source.cpp
```diff
@@ -34,28 +34,26 @@ std::string const& StreamingDataSource::Identity() const {
 }
 
 StreamingDataSource::StreamingDataSource(
```
cwaldren-ld, line 36: Follows from re-arranging constructor args.

libs/server-sdk/src/data_systems/background_sync/sources/streaming/streaming_data_source.hpp
```diff
@@ -6,50 +6,44 @@
...
 
-    void Init(std::optional<data_model::SDKDataSet> initial_data) override;
-    void StartAsync() override;
+    void StartAsync(data_interfaces::IDestination* dest,
+                    data_model::SDKDataSet const* bootstrap_data) override;
     void ShutdownAsync(std::function<void()> completion) override;
 
     [[nodiscard]] std::string const& Identity() const override;
 
    private:
```
cwaldren-ld, line 35: To match the constructor args, the members are re-arranged (so the order of initialization is visually correct.)

libs/server-sdk/src/data_systems/background_sync/sources/polling/polling_data_source.hpp
```diff
@@ -38,19 +38,35 @@ class PollingDataSource
     void DoPoll();
     void HandlePollResult(network::HttpResult const& res);
 
+    Logger const& logger_;
+
+    // Status manager is used to report the status of the data source. It must
+    // outlive the source. This source performs asynchronous
+    // operations, so a completion handler might invoke the status manager after
+    // the it has been destroyed.
```
keelerm84, line 46:
```suggestion
    // operations, so a completion handler might invoke the status manager after
    // it has been destroyed.
```

keelerm84 (approved):

PR #309, opened 2023-11-27
feat: omit empty items from data model JSON serialization

cwaldren-ld:
The server-side SDK is able to treat missing JSON fields as empty ones in-memory. 

For example, in a segment if `included` was missing then the SDK infers it's an empty array. 

This change makes the SDK write out data using that same logic. If an in-memory field is conceptually empty, then omit it from the written data. 

This is an optimization, so I haven't audited all data models to enforce this behavior - the places I've updated are things I'm using in unit tests where I'd like to omit missing (irrelevant for the test) fields.


cwaldren-ld:
Mac CI is expected to fail.

kinyoklion (approved):

PR #310, opened 2023-11-28
feat: flesh out lazy load system

cwaldren-ld:
This implements the Lazy Load system, which is 1 of 2 possible Data Systems in the server-side SDK. 

It's a caching layer over an underlying `ISerializedDataReader`. When items are requested by the SDK, it first checks to see if the item is fresh or stale; if stale, it queries the reader.


libs/server-sdk/src/data_components/expiration_tracker/expiration_tracker.cpp
```diff
@@ -87,8 +87,9 @@ ExpirationTracker::TrackState ExpirationTracker::State(TimePoint expiration,
 void ExpirationTracker::ScopedTtls::Set(DataKind kind,
                                         std::string const& key,
                                         TimePoint expiration) {
-    data_[static_cast<std::underlying_type_t<DataKind>>(kind)].Data().insert(
-        {key, expiration});
+    data_[static_cast<std::underlying_type_t<DataKind>>(kind)]
+        .Data()
+        .insert_or_assign(key, expiration);
```
cwaldren-ld, line 92: The previous behavior was that if the key already existed, this was a no-op. I think that's incorrect since that means you can't extend (update) the expiry of an item. But perhaps I am holding it wrong.

libs/server-sdk/src/data_components/serialization_adapters/json_deserializer.cpp
```diff
@@ -10,8 +10,8 @@
 namespace launchdarkly::server_side::data_components {
 
 JsonDeserializer::JsonDeserializer(
-    data_interfaces::ISerializedDataReader& reader)
-    : flag_kind_(), segment_kind_(), reader_(reader) {}
+    std::shared_ptr<data_interfaces::ISerializedDataReader> reader)
+    : flag_kind_(), segment_kind_(), reader_(std::move(reader)) {}
```
cwaldren-ld, line 14: `ISerializedDataReader` is shared ownership because it's constructed outside the SDK. 

Previously, the `JsonDeserializer` held a reference to it and the Lazy Load system held the shared_ptr - but it's simpler to just have the deserializer own it. That way, the Lazy Load system only needs to own one thing instead of two.

libs/server-sdk/src/data_components/serialization_adapters/json_deserializer.cpp
```diff
@@ -36,7 +36,11 @@ JsonDeserializer::AllSegments() const {
 }
 
 std::string const& JsonDeserializer::Identity() const {
-    return reader_.Identity();
+    return reader_->Identity();
+}
+
+bool JsonDeserializer::Initialized() const {
+    return reader_->Initialized();
 }
```
cwaldren-ld, line 44: Since `JsonDeserializer` now mirrors all the functions of `ISerializedDataReader` (`Initialized` was missing), we can simply have it own the `shared_ptr<ISerializedDataReader>` instead of taking a reference.

This way the Lazy Load system only needs to own one reader instead of the two (to get at `Initialized`).

libs/server-sdk/src/data_systems/lazy_load/lazy_load_system.cpp
```diff
@@ -20,15 +28,17 @@ void LazyLoad::Shutdown() {}
 
 std::shared_ptr<data_model::FlagDescriptor> LazyLoad::GetFlag(
     std::string const& key) const {
-    auto const state = tracker_.State(Keys::kAllSegments, time_());
+    auto const state =
+        tracker_.State(data_components::DataKind::kFlag, key, time_());
```
cwaldren-ld, line 32: todo: I thought this was a typo (using `kAllSegments` here) - am I using it wrong?

libs/server-sdk/src/data_systems/lazy_load/lazy_load_system.cpp
```diff
@@ -20,15 +28,17 @@ void LazyLoad::Shutdown() {}
...
         state, [this, &key]() { RefreshFlag(key); },
         [this, &key]() { return cache_.GetFlag(key); });
 }
 
 std::shared_ptr<data_model::SegmentDescriptor> LazyLoad::GetSegment(
     std::string const& key) const {
-    auto const state = tracker_.State(Keys::kAllSegments, time_());
+    auto const state =
+        tracker_.State(data_components::DataKind::kSegment, key, time_());
     return Get<std::shared_ptr<data_model::SegmentDescriptor>>(
```
cwaldren-ld, line 42: todo: Here I thought another typo - should be checking the state of that particular segment (?)

libs/server-sdk/tests/lazy_load_system_test.cpp
```diff
@@ -0,0 +1,235 @@
...
+
+TEST_F(LazyLoadTest, AllFlagsRefreshesIndividualFlagExpiration) {
+    using TimePoint = data_systems::LazyLoad::ClockType::time_point;
+
+    constexpr auto refresh_ttl = std::chrono::seconds(10);
+
+    built::LazyLoadConfig const config{
+        built::LazyLoadConfig::EvictionPolicy::Disabled, refresh_ttl,
+        mock_reader};
+
```
cwaldren-ld, line 216: TODO: this test isn't implemented yet.

libs/server-sdk/include/launchdarkly/server_side/integrations/iserialized_item_kind.hpp
```diff
@@ -51,4 +52,8 @@ inline bool operator==(ISerializedItemKind const& lhs,
     return lhs.Namespace() == rhs.Namespace();
 }
```
cwaldren-ld, line 54: For googletest.

libs/server-sdk/src/client_impl.cpp
```diff
@@ -116,80 +116,59 @@ ClientImpl::ClientImpl(Config config, std::string const& version)
     run_thread_ = std::move(std::thread([&]() { ioc_.run(); }));
```
cwaldren-ld, line 116: The changes made here are a stopgap to get the basic background sync + streaming use-case working (call `StartAsync`, check the future to see if the SDK is initialized.)

In the DS spec, I've tried to solve the problem of separating out the state of the system and the state of the data. They are currently intertwined. 

For example, once we have initial data, we want to say that _data is available_ (forever) - but the current data source status might become _interrupted_, so it's not a good proxy for that. 

Here when the source status changes, we ignore the new state and instead check the result of `data_system->Initialized()` to see if data is available. So a sequence of events might look like:

```
Status::kInitializing
data_system->Initialized() = false

Status::kInterrupted 
data_system->Initialized() = false

Status::kValid 
data_system->Initialized() = true

Status::kInterrupted
data_system->Initialized() = true
```


In future revisions, we should put the data state into the Status Manager so that we can support this in a first class way. There'd be no need for a `data_system->Initialized()` method because we could check the status manager instead as the source of truth.

libs/server-sdk/tests/server_c_bindings_test.cpp
```diff
@@ -52,9 +52,11 @@ TEST(ClientBindings, RegisterDataSourceStatusChangeListener) {
     LDListenerConnection connection =
         LDServerSDK_DataSourceStatus_OnStatusChange(sdk, listener);
```
cwaldren-ld, line 54: This change is an implication of the DataSourceStatus/Initialized changes I commented above. 

If the SDK is in "offline" mode, that means the data system + events are disabled. The SDK isn't making any external connections, to a database or LaunchDarkly.

libs/server-sdk/src/data_components/dependency_tracker/data_kind.hpp
```diff
@@ -1,7 +1,26 @@
 #pragma once
 
+#include <launchdarkly/detail/unreachable.hpp>
+
 #include <cstddef>
+#include <ostream>
 
 namespace launchdarkly::server_side::data_components {
 enum class DataKind : std::size_t { kFlag = 0, kSegment = 1, kKindCount = 2 };
+
```
cwaldren-ld, line 10: For log messages.

libs/server-sdk/src/data_components/serialization_adapters/json_deserializer.cpp
```diff
@@ -10,33 +10,40 @@
 namespace launchdarkly::server_side::data_components {
 
 JsonDeserializer::JsonDeserializer(
```
cwaldren-ld, line 12: In the header, I refactored `Deserialize` to be a common base for deserializing a single item or a collection of items.

libs/server-sdk/src/data_interfaces/system/idata_system.hpp
```diff
@@ -21,19 +21,13 @@ class IDataSystem : public IStore {
      */
     virtual void Initialize() = 0;
```
cwaldren-ld, line 23: Removed Shutdown because it has no use at this time. Shutting down the `io_service` causes all of our async stuff to close, and then destructors take care of the rest.

PR #329, opened 2023-12-12
ci: simulate mac + windows releases, fix openssl root path

cwaldren-ld:


.github/actions/sdk-release/action.yml
```diff
@@ -92,28 +95,12 @@ runs:
       if: runner.os == 'Windows'
       uses: ilammy/msvc-dev-cmd@v1
 
-    - name: Upgrade OpenSSL
```
cwaldren-ld, line 95: All this can be removed since we have the install-openssl step.

kinyoklion (approved):

PR #330, opened 2023-12-12
refactor: tombstone deserialization

cwaldren-ld:
This cleans up some nasty tombstone deserialization code in the initial server-side release. It swaps from an ad-hoc deserializer to one compatible with `boost::value_to`.

I've also removed the `StorageItem` concept that I had introduced to make it clear that a deserialized item was a tombstone or not. That's already taken care of by `ItemDescriptor` - I just wanted to see if a more explicit data structure could work. I think it's not worth the trouble at the moment.


libs/client-sdk/src/data_sources/data_source_event_handler.cpp
```diff
@@ -92,7 +92,7 @@ DataSourceEventHandler::MessageStatus DataSourceEventHandler::HandleMessage(
             status_manager_.SetError(
                 DataSourceStatus::ErrorInfo::ErrorKind::kInvalidData,
                 kErrorParsingPut);
```
cwaldren-ld, line 94: Removing some redundant namespace qualifiers while I'm here.

libs/client-sdk/src/data_sources/data_source_event_handler.cpp
```diff
@@ -145,22 +144,22 @@ DataSourceEventHandler::MessageStatus DataSourceEventHandler::HandleMessage(
...
-        auto res = boost::json::value_to<
-            tl::expected<DataSourceEventHandler::DeleteData, JsonError>>(
+        auto res = boost::json::value_to<tl::expected<DeleteData, JsonError>>(
             boost::json::parse(data));
         if (res.has_value()) {
-            handler_.Upsert(context_, res.value().key,
-                            ItemDescriptor(res.value().version));
-            return DataSourceEventHandler::MessageStatus::kMessageHandled;
+            handler_.Upsert(
+                context_, res.value().key,
```
cwaldren-ld, line 153: Explicit construction of the tombstone strong typedef is now needed since the constructor only accepts those instead of raw `uint64_t`.

libs/server-sdk/src/data_systems/lazy_load/lazy_load_system.hpp
```diff
@@ -110,15 +110,7 @@ class LazyLoad final : public data_interfaces::IDataSystem {
             status_manager_.SetState(DataSourceState::kValid);
 
             if (auto optional_item = *expected_item) {
```
cwaldren-ld, line 112: Have the transformation from `SerializedItemDescriptor -> StorageItem -> ItemDescriptor` was unnecessarily complex. Now it is back to just `SerializedItemDescriptor -> ItemDescriptor`.

libs/server-sdk/src/data_interfaces/source/idata_reader.hpp
```diff
@@ -20,30 +20,17 @@ namespace launchdarkly::server_side::data_interfaces {
...
-        if (std::holds_alternative<Tombstone>(item)) {
-            return data_model::ItemDescriptor<T>(std::get<Tombstone>(item));
-        }
-        return data_model::ItemDescriptor<T>(std::move(std::get<T>(item)));
-    }
-
     using Error = std::string;
 
     template <typename T>
-    using Single = std::optional<StorageItem<T>>;
```
cwaldren-ld, line 40: Eliminating the `StorageItem` concept because it introduced too much complexity. If variants weren't a pain to deal with we could probably totally switch to using it - but for now, let's just keep using `ItemDescriptor` which is a sort of pseudo-variant. 

Keeping the strong typedef for tombstone because that was helpful for readability.

kinyoklion (approved):

PR #332, opened 2023-12-12
ci: use launchdarkly publish docs action

cwaldren-ld:


.github/actions/publish-docs/action.yml
```diff
@@ -5,18 +5,15 @@ inputs:
...
-        path: gh-pages
-        ref: gh-pages
-    - name: Publish Docs
-      shell: bash
-      run: |
-        echo "Publishing docs for: $WORKSPACE"
-        ./scripts/publish-doc.sh $WORKSPACE
-      env:
-        WORKSPACE: ${{ inputs.workspace_path }}
+        docs_path: ${{ inputs.workspace_path }}
```
kinyoklion, line 17: I am reasonably sure this should be in a subdirectory. `${{ inputs.workspace_path }}/docs` 

The output in the doxyfile contains `docs` and the old script used `docs`.

kinyoklion (changes requested):

.github/actions/publish-docs/action.yml
cwaldren-ld: I did that originally (first commit), and it then broke all existing docs links (like from README + LaunchDarkly public docs).

kinyoklion: Not including it means you are publishing all the source into docs which happens to include docs. If you want it in a sub-directory on the output, then put that into the output path as well.

kinyoklion: image

kinyoklion: What it used to, and correctly, looked like:
image

kinyoklion: So, before we took the $workspace/docs and put them into $workspace/docs.

Currently you are putting $workspace/ which happens to contain docs into $workspace/.

kinyoklion: Is this what you tried?
```
docs_path: ${{ inputs.workspace_path }}/docs
output_path: ${{ inputs.workspace_path }}/docs
```

If it doesn't work, then we need to fix a bug in the action.

cwaldren-ld: Ok, I had only added `/docs` to the `docs_path`. That was the problem. Updated both and all is working now (+ I've fixed up the `gh-pages` branch.)

PR #342, opened 2023-12-14
feat: allow LDStatus to carry an arbitrary string message

cwaldren-ld:
Our C bindings return `LDStatus`, which wraps an `Error` enum. This is an opaque type so that we could modify how the error was represented.

That time has come, and I've modified the underlying `Error` type to be a `variant<ErrorCode, string>`. This supports reporting arbitrary string messages. The immediate use case is reporting errors from the Redis integration, which doesn't use status codes.

### DIFF
--- a/libs/common/include/launchdarkly/bindings/c/status.h
+++ b/libs/common/include/launchdarkly/bindings/c/status.h
@@ -7,37 +7,38 @@
 #include <stdbool.h>
 
 #ifdef __cplusplus
-extern "C" {  // only need to export C interface if
+extern "C" {
+// only need to export C interface if
 // used by C++ source code
 #endif
 
 typedef struct _LDStatus* LDStatus;
 
 /**
- * Returns a string representing the error stored in an LDStatus, or
- * NULL if the result indicates success.
- * @param error Result to inspect.
- * @return String or NULL. The returned string is valid only while the LDStatus
+ * Returns a string describing the error reported by the LDStatus, or
+ * NULL if the status indicates success.
+ * @param status Status to inspect.
+ * @return String or NULL. The string is valid only while the LDStatus
  * is alive.
  */
-LD_EXPORT(char const*) LDStatus_Error(LDStatus res);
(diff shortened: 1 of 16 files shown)


libs/common/include/launchdarkly/bindings/c/status.h
```diff
@@ -7,37 +7,38 @@
 #include <stdbool.h>
```
cwaldren-ld, line 8: Doc wording + parameter renamings to more consistently call this a "status".

libs/common/include/launchdarkly/detail/c_binding_helpers.hpp
```diff
@@ -9,21 +9,25 @@
```
cwaldren-ld, line 9: Formatting changes only.

libs/common/include/launchdarkly/error.hpp
```diff
@@ -27,7 +26,11 @@ enum class Error : std::uint32_t {
     kMax = std::numeric_limits<std::uint32_t>::max()
 };
 
-char const* ErrorToString(Error err);
```
cwaldren-ld, line 30: Made `Error` be a variant, and renamed the existing `Error` (enum) to `ErrorCode`. 

I also added a `bool operator==(Error, ErrorCode)` to satisfy existing tests.

libs/common/src/bindings/c/status.cpp
```diff
@@ -7,27 +7,27 @@
 
 using namespace launchdarkly;
 
-#define TO_ERROR(ptr) (reinterpret_cast<Error*>(ptr))
```
cwaldren-ld, line 10: Renamings to better reflect "status", + const correctness.

libs/common/include/launchdarkly/error.hpp
kinyoklion: I am confused. LDStatus is Opaque, but Error, in C++ land, was not?

PR #345, opened 2023-12-15
feat: redis data source C bindings

cwaldren-ld:
This adds C bindings for creating a Redis Source, as well as a LazyLoad config method to accept pointers to the C type representing an instantiated Redis Source.

Demo:
```c
LDServerLazyLoadBuilder lazy_builder = LDServerLazyLoadBuilder_New();

struct LDServerLazyLoadRedisResult out_result;
if (!LDServerLazyLoadRedisSource_New(uri, prefix, &out_result))
     printf("%s\n", out_result.error_message);
     exit(1);
}

LDServerLazyLoadBuilder_SourcePtr(lazy_builder, out_result.source);

LDServerConfigBuilder_DataSystem_LazyLoad(cfg_builder, lazy_builder);
```

### DIFF
--- a/libs/server-sdk-redis-source/src/bindings/redis/redis_source.cpp
+++ b/libs/server-sdk-redis-source/src/bindings/redis/redis_source.cpp
@@ -12,7 +12,7 @@ using namespace launchdarkly::server_side::integrations;
 LD_EXPORT(bool)
 LDServerLazyLoadRedisSource_New(char const* uri,
                                 char const* prefix,
-                                LDServerLazyLoadResult* out_result) {
+                                LDServerLazyLoadRedisResult* out_result) {
     LD_ASSERT_NOT_NULL(uri);
     LD_ASSERT_NOT_NULL(prefix);
     LD_ASSERT_NOT_NULL(out_result);
@@ -20,7 +20,7 @@ LDServerLazyLoadRedisSource_New(char const* uri,
     // Explicitely zero out the exception_msg buffer in case the exception is
     // shorter than the buffer.
     memset(out_result->error_message, 0,
-           sizeof(LDServerLazyLoadResult::error_message));
+           sizeof(LDServerLazyLoadRedisResult::error_message));
 
     // Ensure the source pointer isn't garbage.
     out_result->source = nullptr;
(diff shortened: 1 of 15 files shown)


libs/server-sdk-redis-source/src/bindings/redis/redis_source.cpp
```diff
@@ -0,0 +1,52 @@
...
+
+LD_EXPORT(bool)
+LDServerLazyLoadRedisSource_New(char const* uri,
+                                char const* prefix,
+                                LDServerLazyLoadRedisResult* out_result) {
+    LD_ASSERT_NOT_NULL(uri);
+    LD_ASSERT_NOT_NULL(prefix);
+    LD_ASSERT_NOT_NULL(out_result);
+
+    // Explicitely zero out the exception_msg buffer in case the exception is
```
cwaldren-ld, line 20:
```suggestion
    // Explicitly zero out the exception_msg buffer in case the exception is
```

libs/server-sdk/src/bindings/c/builder.cpp
```diff
@@ -236,6 +255,45 @@ LDServerDataSourcePollBuilder_Free(LDServerDataSourcePollBuilder b) {
...
+LDServerLazyLoadBuilder_New() {
+    return FROM_LAZYLOAD_BUILDER(new DataSystemBuilder::LazyLoad());
+}
+
+LD_EXPORT(void)
+LDServerLazyLoadBuilder_Free(LDServerLazyLoadBuilder b) {
+    delete TO_LAZYLOAD_BUILDER(b);
+}
+
+LD_EXPORT(void)
```
cwaldren-ld, line 268: Wasn't entirely sure what to do here. Normally we'd pass in a struct with some function pointers, to allow the C users to truly implement the interface - but that would require a ton of wrappers for the various return types, and I don't think it's necessary for a first pass.

Instead, I want the RedisSource to return a C++ implementation of the interface, which we pass into the LazyLoad builder. 

This does mean the contract is entirely in documentation, there's no type safety here.

kinyoklion (approved):

PR #362, opened 2024-01-19
feat: always inline contexts for feature events

keelerm84:


libs/internal/src/events/asio_event_processor.cpp
```diff
@@ -255,7 +255,7 @@ std::vector<OutputEvent> AsioEventProcessor<SDK>::Process(
 
                 if (event.require_full_event) {
                     out.emplace_back(FeatureEvent{std::move(base),
-                                                  event.context.KindsToKeys()});
```
cwaldren-ld, line 258: To fix the formatting errors, you can run clang format (maybe part of clang package on Arch?)

cwaldren-ld (approved):

PR #363, opened 2024-01-22
feat: redact anonymous attributes within feature events

keelerm84:


libs/internal/include/launchdarkly/context_filter.hpp
```diff
@@ -34,6 +34,17 @@ class ContextFilter {
...
+    /**
+     * Filter the given context and produce a JSON value. If a provided context
+     * is anonymous, all attributes are redacted.
+     *
+     * Only call this method for valid contexts.
+     *
+     * @param context The context to redact.
+     * @return JSON suitable for an analytics event.
+     */
+    JsonValue filter_with_anonymous_redaction(Context const& context);
```
keelerm84, line 46: Wasn't sure what the casing convention was for this repo so don't hesitate to nitpick!

libs/internal/include/launchdarkly/context_filter.hpp
```diff
@@ -48,6 +59,15 @@ class ContextFilter {
...
     };
 
+    /**
+     * Internal delegate function for filter and filter_with_anonymous_redaction.
+     *
+     * @param context The context to redact.
+     * @param redactAnonymous Whether to redact all attributes from anonymous contexts.
+     * @return JSON suitable for an analytics event.
+     */
+    JsonValue filter_(Context const& context, bool redactAnonymous);
```
keelerm84, line 69: Trailing `_` the preference? Or something like `internalFilter`?

cwaldren-ld: Should be able to call it just `filter`, as an overload

libs/internal/tests/context_filter_test.cpp
```diff
@@ -173,3 +173,56 @@ TEST(ContextFilterTests, AllAttributesPrivateMultiContext) {
...
+            .Name("Bob")
+            .Anonymous(true)
+            .Set("email", "email.email@email")
+            .Set("array", {false, true, "bacon"})
+            .Set("object", Object{{"test", true}, {"second", false}})
+            .SetPrivate("isCat", false)
+            .AddPrivateAttribute("/object/test")
+            .Kind("org", "org-key")
+            .Set("isCat", true)
+            .Set("email", "cat@cat.cat")
```
cwaldren-ld, line 215: TIL that `cat` is a real TLD (Catalan): https://www.gandi.net/en-US/domain/tld/cat

cwaldren-ld (approved):

PR #364, opened 2024-01-24
refactor: Update context filter class to match preferred coding style

keelerm84:


keelerm84:
Feedback from https://github.com/launchdarkly/cpp-sdks/pull/363 mentioned that this file wasn't following the newly established conventions. So I'm fixing that in a separate PR.

There is one method on the context filter class, `emplace`, which I didn't convert to PascalCase. I see that `emplace` is a more common method name in C++, so I thought you may want to keep that consistent. Happy to change it if not.

cwaldren-ld:
Let's have `emplace` use PascalCase as well.

cwaldren-ld (approved):
(approved pending renaming -> `Emplace`)

PR #368, opened 2024-02-07
fix: handle omitted evaluation result value when deserializing client-side JSON payload

cwaldren-ld:
**Background**
When the client-side SDK is deserializing an evaluation result, it may encounter a `null` evaluation value for a particular flag. That is, the value is literally `null` in JSON.

It might also receive an evaliuation result with an _omitted_ value. Some server-side SDKs may produce this because they don't treat null/missing as equivalent. 

**Problem**
In that case, the client-side SDK would throw a schema error. This isn't correct since producing a null/omitted value is legitimate. 

**Solution**
The deserialization code now takes into account the equivalence of null/omission and doesn't throw an error.

### DIFF
--- a/libs/internal/src/serialization/json_evaluation_result.cpp
+++ b/libs/internal/src/serialization/json_evaluation_result.cpp
@@ -8,7 +8,6 @@
 #include <boost/json.hpp>
 
 namespace launchdarkly {
-
 tl::expected<std::optional<EvaluationResult>, JsonError> tag_invoke(
     boost::json::value_to_tag<
         tl::expected<std::optional<EvaluationResult>, JsonError>> const& unused,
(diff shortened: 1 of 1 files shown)


kinyoklion (approved):

cwaldren-ld:
Did a more deep audit of the code. I'm seeing an (un-contract-tested) problem in the handling of variations where `null` evaluation result will result in entering the type-check branch and return a `WRONG_TYPE` error. 

That's not correct. So this PR shouldn't be merged yet until I make some contract tests that cover this (and fix it.)

cwaldren-ld:
I'm going to merge this as it's a good defensive change in case the SDK is sent bad/malformed data. I've created a ticket to investigate whether a `null` flag value is possible or not, and if we need to have contract tests to cover this.

Issue #373, opened 2024-03-10
BoolVariation is not thread safe


magnushakansson:
**Is this a support request?**
No

**Describe the bug**
When BoolVariation is called on multiple threads, there is a segv.

**To reproduce**
I have modified the example program in https://github.com/launchdarkly/cpp-sdks/blob/main/examples/hello-cpp-server/main.cpp.
Replace lines https://github.com/launchdarkly/cpp-sdks/blob/main/examples/hello-cpp-server/main.cpp#L58C1-L66C1 with:
```
     auto LoopGetFeatures = [&client] {
        while (true) {
            auto const context = ContextBuilder().Kind("user", "example-user-key").Name("Sandy").Build();
            bool const flag_value = client.BoolVariation(context, FEATURE_FLAG_KEY, false);
        }
    };

    std::thread t1(LoopGetFeatures);
    std::thread t2(LoopGetFeatures);
    std::thread t3(LoopGetFeatures);
    std::thread t4(LoopGetFeatures);

    t1.join();
    t2.join();
    t3.join();
    t4.join();
```
Make sure that `FEATURE_FLAG_KEY` is set in the backend you use.

**Expected behavior**
The example program shall run without problem.

**Logs**
Before the program crashes, you will see a number of
```
[LaunchDarkly] Invalid flag configuration detected in flag "(no parent)": prerequisite relationship to "GLOBAL_FLAG_TRUE" caused a circular reference; this is probably a temporary condition due to an incomplete update
```

**SDK version**
3.4.0

**Language version, developer tools**
C++20, Clang 17.

**OS/platform**
Debian 11

**Additional context**
It seems like the two sets in https://github.com/launchdarkly/cpp-sdks/blob/main/libs/server-sdk/src/evaluation/detail/evaluation_stack.hpp#L56C1-L57C52 lack protection.
When I add a mutex for them, the crash is avoided but the log line mentioned above is still there so something more seems to be needed.

cwaldren-ld:
Hi @magnushakansson , you are correct that it is not thread-safe, and I apologize for having to discover it yourself rather than read it in documentation. That was an oversight.

We'd like to get this properly supported so that it's safe to call from multiple threads in parallel. This involves creating a new `EvaluationStack` for each call, rather than sharing one. 

Would you mind checking out this branch and re-testing the scenario on your machine: `cw/sc-235941/thread-safe-evaluation`? 

On my machine, it no longer crashes.

[...]

PR #374, opened 2024-03-11
fix: Evaluate should not share EvaluationStack between calls

cwaldren-ld:
Addresses #373.

### DIFF
--- a/libs/server-sdk/src/evaluation/evaluator.hpp
+++ b/libs/server-sdk/src/evaluation/evaluator.hpp
@@ -40,7 +40,7 @@ class Evaluator {
 
     /**
      * Evaluates a flag for a given context. Does not record prerequisite
-     * events. Warning: not thread safe.
+     * events.
      *
      * @param flag The flag to evaluate.
      * @param context The context to evaluate the flag against.
--- a/libs/server-sdk/src/evaluation/evaluator.hpp
+++ b/libs/server-sdk/src/evaluation/evaluator.hpp
@@ -36,7 +36,6 @@ class Evaluator {
     [[nodiscard]] EvaluationDetail<Value> Evaluate(
         data_model::Flag const& flag,
         Context const& context,
-        EvaluationStack& stack,
         EventScope const& event_scope);
 
     /**
--- a/libs/server-sdk/src/evaluation/evaluator.hpp
+++ b/libs/server-sdk/src/evaluation/evaluator.hpp
@@ -6,33 +6,37 @@
 #include <launchdarkly/logging/logger.hpp>
 #include <launchdarkly/value.hpp>
 
-#include "bucketing.hpp"
 #include "detail/evaluation_stack.hpp"
 #include "evaluation_error.hpp"
 
 #include "../data_interfaces/store/istore.hpp"
 #include "../events/event_scope.hpp"
 
-#include <tl/expected.hpp>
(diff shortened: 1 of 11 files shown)


cwaldren-ld:
There are additional thread-safety concerns I've discovered while testing this. Specifically, although we may be able to make the evaluator thread safe, we need to ensure the data system implementations are thread safe as well.

Specific notes:
- DataSystem::Initialized() needs to to be thread safe
- If lazy loading, refreshing needs to be thread safe (the readers are supposed to be by definition, but the expiration tracker needs to be)
- If background sync, the memory store is already thread safe.

libs/server-sdk/tests/evaluator_tests.cpp
```diff
@@ -246,3 +249,30 @@ TEST_F(EvaluatorTests, FlagWithExperimentTargetingMissingContext) {
     ASSERT_EQ(*detail, Value(false));
     ASSERT_EQ(detail.Reason(), EvaluationReason::Fallthrough(false));
 }
+
+TEST_F(EvaluatorTests, ThreadSafeEvaluation) {
```
cwaldren-ld, line 253: This type of test is what I've done in the past, and I'm not really sure of a better way to do it. 

Basically spin up a bunch of threads and see if ASAN is triggered.

libs/server-sdk/src/evaluation/evaluator.hpp
```diff
@@ -6,33 +6,37 @@
...
      * @param flag The flag to evaluate.
      * @param context The context to evaluate the flag against.
+     * @param stack The evaluation stack used for detecting circular references.
      * @param event_scope The event scope used for recording prerequisite
      * events.
      */
     [[nodiscard]] EvaluationDetail<Value> Evaluate(
         data_model::Flag const& flag,
         Context const& context,
+        EvaluationStack& stack,
```
kinyoklion, line 39: Is there a reason the public API needs to take a stack versus making one. Or are we re-using the public version internal to the evaluator as well?

libs/server-sdk/src/evaluation/evaluator.hpp
```diff
@@ -6,33 +6,37 @@
...
         EventScope const& event_scope);
 
     /**
```
kinyoklion, line 42: This next function as a "not thread safe" warning as well.

libs/server-sdk/src/evaluation/evaluator.hpp
kinyoklion: We also may want to consider a more generic `EvaluationState` that contains the evaluation stack. 

Eventually you have big segment status and big segment membership that persist the duration of an evaluation.

kinyoklion (approved):

libs/server-sdk/src/evaluation/evaluator.hpp
cwaldren-ld: Yes, we are using this in AllFlags. So we setup the stack object once, then run all the evaluations on it. 

I think I did that because it felt bad to be constructing the stack over and over for each flag - but I didn't have any data, so I'd be fine pushing it back into the Evaluator. 

It'd be safer that way too, if the stack somehow got un-balanced by a failed evaluation.

kinyoklion (approved):

PR #381, opened 2024-03-15
ci: upgrade googletest

cwaldren-ld:
This upgrades googletest to a much more recent commit. The intention is to see if it helps with the `DEADLYSIGNAL` issue that has been plaguing this repo's CI for the last couple weeks. 

Underlying theory is that possibly there is some incompatibilitiy between the old googletest and gcc.


cwaldren-ld:
So, looks like there was one failure where the runner ran for 1+ hour out of 5 runs in the `build-test-sse` job (I've been using that one to trigger the DEADLYSIGNAL.)

 This could be a total coincidence, but it does seem like a better failure rate than 1 out of 2 or 3.

kinyoklion (approved):

PR #386, opened 2024-04-02
build: bumping slsa-github-generator version to resolve issue with retrieving Rekor public keys

rsoberano-ld:
Provenance generation fails due to Rekor public key errors were identified as a known issue and fixed in version 1.10.0 per: https://github.com/slsa-framework/slsa-github-generator/issues/3350

### DIFF
--- a/.github/workflows/manual-sdk-release-artifacts.yml
+++ b/.github/workflows/manual-sdk-release-artifacts.yml
@@ -67,7 +67,7 @@ jobs:
       actions: read
       id-token: write
       contents: write
-    uses: slsa-framework/slsa-github-generator/.github/workflows/generator_generic_slsa3.yml@v1.7.0
+    uses: slsa-framework/slsa-github-generator/.github/workflows/generator_generic_slsa3.yml@v1.10.0
     with:
       base64-subjects: "${{ needs.release-sdk.outputs[format('hashes-{0}', matrix.os)] }}"
       upload-assets: true
--- a/.github/workflows/release-please.yml
+++ b/.github/workflows/release-please.yml
@@ -102,7 +102,7 @@ jobs:
       actions: read
       id-token: write
       contents: write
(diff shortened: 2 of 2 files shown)


kinyoklion (commented):
Should we consider rennovate for this?

https://github.com/slsa-framework/slsa-github-generator/blob/main/RENOVATE.md

rsoberano-ld:
> Should we consider rennovate for this?
> 
> https://github.com/slsa-framework/slsa-github-generator/blob/main/RENOVATE.md

I'm a little confused - this snippet would setup a Renovate exception to allow us to use tags for `slsa-github-generator` if Renovate was blocking tags in favor of digests previously, which I don't think is what is going on here(?)

kinyoklion:
> > Should we consider rennovate for this?
> > https://github.com/slsa-framework/slsa-github-generator/blob/main/RENOVATE.md
> 
> I'm a little confused - this snippet would setup a Renovate exception to allow us to use tags for `slsa-github-generator` if Renovate was blocking tags in favor of digests previously, which I don't think is what is going on here(?)

It is their best practice for using rennovate, we could also do it without that. It just seems reasonable to automate it.

kinyoklion (approved):

rsoberano-ld:
> > > Should we consider rennovate for this?
> > > https://github.com/slsa-framework/slsa-github-generator/blob/main/RENOVATE.md
> > 
> > 
> > I'm a little confused - this snippet would setup a Renovate exception to allow us to use tags for `slsa-github-generator` if Renovate was blocking tags in favor of digests previously, which I don't think is what is going on here(?)
> 
> It is their best practice for using rennovate, we could also do it without that. It just seems reasonable to automate it.

Ahh I see what you mean - I do agree having Renovate in general would be helpful here to help keep us up to date, especially when it comes to bumping up these versions across all of our SDK repos.

PR #389, opened 2024-04-05
fix: handle service endpoints with custom ports correctly

cwaldren-ld:


libs/server-sent-events/src/client.cpp
```diff
@@ -589,11 +589,22 @@ std::shared_ptr<Client> Builder::build() {
...
+            return nullptr;
+        }
+    }
+
+    // The resolver accepts either a port number or a service name. If the
+    // URL specifies a port, use that - otherwise, pass in the scheme as the
+    // service name (which will be either http or https due to the check
+    // above.)
     std::string service = uri_components->has_port() ? uri_components->port()
                                                      : uri_components->scheme();
```
cwaldren-ld, line 604: The issue here was that if a port was provided, say 9090, that's `!= "https"` so it'd end up being a plain http connection.

kinyoklion: Does this affect any other connections? Events, polling?

cwaldren-ld: Looks like it affected our AsioRequester which is used by Events and Polling. I've updated it there as well.

libs/internal/include/launchdarkly/network/asio_requester.hpp
```diff
@@ -323,7 +323,7 @@ class AsioRequester {
                 request->Port().value_or(request->Https() ? "https" : "http");
 
             std::shared_ptr<ssl::context> ssl;
```
cwaldren-ld, line 325: `service` might be something like `9090`, but we still want to use HTTPs if that was the scheme.

kinyoklion (approved):

PR #396, opened 2024-04-29
feat: deprecate StartAsync

cwaldren-ld:


libs/client-sdk/include/launchdarkly/client_side/client.hpp
```diff
@@ -36,6 +45,52 @@ class IClient {
...
+     * default values or cached flag values if a persistent store is in use.
+     *
+     * To block until the connection is established, call WaitForInitialization
+     * after calling Initialize.
+     *
+     * The client may still post events to LaunchDarkly even if Initialize isn't
+     * called.
+     *
+     * It is not safe to call Initialize more than once. If the client
+     * encounters an unrecoverable error, destroy it and create a new client.
```
cwaldren-ld, line 60: To actually follow this instruction, the user would need to check the data source status. In reality, people usually just call `exit()` if the waiting fails and have the app restart instead.

PR #399, opened 2024-05-08
feat: add ability to skip TLS peer verification

cwaldren-ld:
This PR adds in a new configuration option allowing users to disable peer verification in TLS handshake. 

Additionally, it implements two new contract test capabilities to verify this behavior (`tls:verify-peer` and `tls:skip-verify-peer`.)

Because the functionality was added to the sse library and the shared asio requester, both the server and client gain this functionality. Therefore I've added bindings for both along with the contract test implementations.


libs/client-sdk/src/data_sources/polling_data_source.cpp
```diff
@@ -88,6 +88,10 @@ PollingDataSource::PollingDataSource(
     auto const& polling_config = std::get<
         config::shared::built::PollingConfig<config::shared::ClientSDK>>(
         data_source_config.method);
+    if (http_properties.Tls().VerifyMode() ==
```
cwaldren-ld, line 91: Could pass a logger dep to the `requester_` and have it do the logging instead. In any case, I just wanted to be sure people see a log message since this is pretty critical. Perhaps it should be at info or warn level.

libs/common/include/launchdarkly/config/shared/builders/http_properties_builder.hpp
```diff
@@ -10,6 +10,45 @@
 
 namespace launchdarkly::config::shared::builders {
 
+/**
+ * Class used for building TLS options used within HttpProperties.
+ * @tparam SDK THe SDK type to build options for. This affects the default
```
keelerm84, line 15:
```suggestion
 * @tparam SDK The SDK type to build options for. This affects the default
```

libs/internal/include/launchdarkly/events/detail/worker_pool.hpp
```diff
@@ -23,17 +24,24 @@ namespace launchdarkly::events::detail {
  */
 class WorkerPool {
    public:
+    WorkerPool(boost::asio::any_io_executor io,
```
keelerm84, line 27: Should this have some documentation?

keelerm84 (approved):

libs/internal/include/launchdarkly/events/detail/worker_pool.hpp
cwaldren-ld: Probably. It's internal but.. yes.

cwaldren-ld: I will add in another PR.

PR #400, opened 2024-05-13
docs: call out indefinite blocking scenerio on StartAsync method

cwaldren-ld:
Since modifying the `StartAsync` method (which returns a future) would be a significant change to the API, we're adding docs instead to call out the "don't block indefinitely" scenario.

### DIFF
--- a/libs/client-sdk/include/launchdarkly/client_side/client.hpp
+++ b/libs/client-sdk/include/launchdarkly/client_side/client.hpp
@@ -28,11 +28,20 @@ class IClient {
 
     /** Connects the client to LaunchDarkly's flag delivery endpoints.
      *
-     * If StartAsync isn't called, the client is able to post events but is
-     * unable to obtain flag data.
+     * StartAsync must be called to obtain fresh flag data.
      *
      * The returned future will resolve to true or false based on the logic
      * outlined on @ref Initialized.
+     *
+     * Blocking indefinitely on the future (e.g. by calling .get() or
+     * .wait()) is highly discouraged. Instead, use a method that takes a
+     * timeout like .wait_for() or .wait_until(), or do not wait.
+     *
+     * Otherwise, the application may hang indefinitely if the client cannot
+     * connect to LaunchDarkly.
+     *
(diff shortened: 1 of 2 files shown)


libs/client-sdk/include/launchdarkly/client_side/client.hpp
```diff
@@ -43,11 +52,16 @@ class IClient {
...
-     * timed out trying to connect to LD, then Initialized returns false (even
-     * if we do have cached flags). If the client connected and got a 401 error,
-     * Initialized is will return false. This serves the purpose of letting the
-     * app know that there was a problem of some kind.
+     * offline mode so there's no need to connect to LaunchDarkly.
+     *
+     * If the client timed out trying to connect to LD, then Initialized returns
+     * false (even if we do have cached flags).
+     *
+     * If the client connected and got a 401 error, Initialized is will return
```
keelerm84, line 60:
```suggestion
     * If the client connected and got a 401 error, Initialized will return
```

libs/server-sdk/include/launchdarkly/server_side/client.hpp
```diff
@@ -42,10 +51,13 @@ class IClient {
...
      * LaunchDarkly and successfully retrieved flags, or 2. it started in
-     * offline mode so there's no need to connect to LaunchDarkly. If the client
-     * timed out trying to connect to LD, then Initialized returns false (even
-     * if we do have cached flags). If the client connected and got a 401 error,
-     * Initialized is will return false. This serves the purpose of letting the
+     * offline mode so there's no need to connect to LaunchDarkly.
+     *
+     * If the client timed out trying to connect to LD, then Initialized returns
+     * false (even if we do have cached flags). If the client connected and
+     * got a 401 error, Initialized is will return false.
```
keelerm84, line 58:
```suggestion
     * got a 401 error, Initialized will return false.
```

libs/client-sdk/include/launchdarkly/client_side/client.hpp
cwaldren-ld: This is will correct.

PR #409, opened 2024-05-28
feat: specify a custom CA file for TLS peer verification

cwaldren-ld:
This adds a new config builder option, `CustomCAFile` and associated C binding to the server and client SDKs.

When specified, the SDK's streaming, polling, and event connections will verify its TLS peer based on the CAs found in this file. The custom file may be un-set by passing an empty string. 

We could instead make empty string be a configuration error, but it would involve changing the signature for the `Build()` methods of the HTTPs properties + TLS properties builders. That might be a safer default, although potentially less convenient.

### DIFF
--- a/libs/client-sdk/src/data_sources/polling_data_source.cpp
+++ b/libs/client-sdk/src/data_sources/polling_data_source.cpp
@@ -74,7 +74,7 @@ PollingDataSource::PollingDataSource(
       status_manager_(status_manager),
       data_source_handler_(
           DataSourceEventHandler(context, handler, logger, status_manager_)),
-      requester_(ioc, http_properties.Tls().PeerVerifyMode()),
+      requester_(ioc, http_properties.Tls()),
       timer_(ioc),
       polling_interval_(
           std::get<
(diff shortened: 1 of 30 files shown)


libs/client-sdk/src/data_sources/polling_data_source.cpp
```diff
@@ -88,6 +88,11 @@ PollingDataSource::PollingDataSource(
     auto const& polling_config = std::get<
         config::shared::built::PollingConfig<config::shared::ClientSDK>>(
         data_source_config.method);
+    if (http_properties.Tls().CustomCAFile()) {
```
cwaldren-ld, line 91: It's not ideal to have these debug level logs scattered in the polling/streaming/event data sources, I think. 

They will need to be kept in sync, and also would be redundant if we had one for the events system.

Instead, this should be done at a higher level, perhaps the top-level config builder. But there's no logger dependency in there, so it'd need a refactoring.

kinyoklion (approved):

PR #415, opened 2024-06-13
fix: ensure LDLogLevel is at least 32 bits wide

cwaldren-ld:


libs/common/include/launchdarkly/bindings/c/logging/log_level.h
```diff
@@ -23,6 +23,8 @@ enum LDLogLevel {
     LD_LOG_INFO = 1,
     LD_LOG_WARN = 2,
     LD_LOG_ERROR = 3,
+    LD_LOG_MAXVAL = 0x7FFFFFFF /* This is not a valid log level, it is used to
```
cwaldren-ld, line 26: Why is it so hard to find confirmation that this actually works on the internet?

libs/common/include/launchdarkly/logging/log_level.hpp
```diff
@@ -8,7 +9,7 @@ namespace launchdarkly {
  * severity. The values must not be changed to ensure backwards compatibility
  * with the C API.
  */
-enum class LogLevel {
+enum class LogLevel : std::uint32_t {
```
cwaldren-ld, line 12: Ensure the C++ underlying type is 32 bits so it can hold whatever the C enum is

libs/common/src/log_level.cpp
```diff
@@ -11,13 +11,12 @@ char const* GetLogLevelName(LogLevel level, char const* default_) {
             return "debug";
         case LogLevel::kInfo:
             return "info";
-        case LogLevel::kWarn: {
```
cwaldren-ld, line 14: This pair of braces was here.. must've been a typo. It shouldn't have caused any issue, but it was breaking the indentation.

kinyoklion (approved):

PR #419, opened 2024-06-26
fix: more helpful error messages for streaming connection failures

cwaldren-ld:
This commit pulls the stringification logic of SSE errors into the SSE package itself, and refactors the enum into a richer sum type using `std::variant`.

### DIFF
--- a/libs/server-sent-events/src/client.cpp
+++ b/libs/server-sent-events/src/client.cpp
@@ -222,7 +222,7 @@ class FoxyClient : public Client,
 
         if (status_class == beast::http::status_class::successful) {
             if (response.result() == beast::http::status::no_content) {
-                errors_(Error::NoContent);
+                errors_(errors::NoContent{});
                 return;
             }
             if (!correct_content_type(response)) {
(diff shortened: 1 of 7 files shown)


libs/server-sent-events/src/error.cpp
```diff
@@ -0,0 +1,61 @@
...
+    out << "timed out reading response body (exceeded " << err.timeout->count()
+        << "ms)";
+    return out;
+}
+
+std::ostream& operator<<(std::ostream& out,
+                         UnrecoverableClientError const& err) {
+    if (err.status == boost::beast::http::status::unauthorized ||
+        err.status == boost::beast::http::status::forbidden) {
+        out << "invalid auth key (HTTP " << static_cast<int>(err.status) << ")";
```
cwaldren-ld, line 35: This could be more explicit (mobile vs SDK key), but that would require some special casing in each SDK and I don't think that logic is worth it. This is clear enough.

libs/server-sent-events/src/client.cpp
```diff
@@ -264,7 +264,7 @@ class FoxyClient : public Client,
                 return async_backoff(backoff_reason(response.result()));
```
kinyoklion, line 264: When does this get logged and what does it look like if it is?

cwaldren-ld: It's what happens when there's a "recoverable" client error and a retry will occur. The log message looks like: 
> "Backing off in (N) seconds due to HTTP status 403 (Forbidden)"

kinyoklion: Cool. Thank you.

kinyoklion (approved):

Issue #411, opened 2024-05-31
hello-cpp-client cannot build because of hidden symbols


PunChaFeng:
**Is this a support request?**
No.
  
  **Describe the bug**
Building hello-cpp-client with the shared C++ client library results in linking errors (Undefined symbols). I think it is because all the C++ symbols are hidden.
  
  **To reproduce**
 - Change cmake options: BUILD_TESTING=OFF; LD_BUILD_SHARED_LIBS=ON
 - Build the project, hello-cpp-client and fails.
 

  **Expected behavior**
The project should be built correctly.

  **Logs**
```
====================[ Build | hello-cpp-client | Debug ]========================
cmake" --build cmake-build-debug --target hello-cpp-client -j 8
[0/2] Re-checking globbed directories...
[103/103] Linking CXX executable examples/hello-cpp-client/hello-cpp-client
FAILED: examples/hello-cpp-client/hello-cpp-client 
: && /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/c++ -g -arch arm64 -isysroot /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX14.5.sdk -Wl,-search_paths_first -Wl,-headerpad_max_install_names  examples/hello-cpp-client/CMakeFiles/hello-cpp-client.dir/main.cpp.o -o examples/hello-cpp-client/hello-cpp-client  -Wl,-rpath,/Users/fengx/Works/cpp-sdks/cmake-build-debug/libs/client-sdk/src  libs/client-sdk/src/liblaunchdarkly-cpp-client.dylib  /opt/homebrew/lib/libboost_system-mt.dylib  /opt/homebrew/lib/libboost_date_time-mt.dylib  -framework CoreFoundation  -framework Security  /opt/homebrew/lib/libboost_system-mt.dylib  /opt/homebrew/lib/libboost_filesystem-mt.dylib  /opt/homebrew/lib/libboost_atomic-mt.dylib  /opt/homebrew/lib/libboost_date_time-mt.dylib  /opt/homebrew/lib/libssl.a  /opt/homebrew/lib/libcrypto.a  /opt/homebrew/lib/libboost_url-mt.dylib  /opt/homebrew/lib/libboost_json-mt.dylib  /opt/homebrew/lib/libboost_container-mt.dylib && :
Undefined symbols for architecture arm64:
[...]

PR #420, opened 2024-07-03
build: add LD_BUILD_EXPORT_ALL_SYMBOLS option to export all symbols

puncha:
Address the issue https://github.com/launchdarkly/cpp-sdks/issues/411.

A new CMake option, LD_BUILD_EXPORT_SYMBOLS, exports all symbols, including CPP and C APIs. It is set to OFF by default.


README.md
```diff
@@ -61,6 +61,7 @@ Various CMake options are available to customize the client/server SDK builds.
 | `LD_BUILD_CONTRACT_TESTS` | Whether the contract test service (used in CI) is built.                                                                                          | Off                                                    | `BUILD_TESTING`                           |
 | `LD_BUILD_EXAMPLES`       | Whether example apps (hello world) are built.                                                                                                     | On                                                     | N/A                                       |
 | `LD_BUILD_SHARED_LIBS`    | Whether the SDKs are built as static or shared libraries.                                                                                         | Off  (static lib)                                      | N/A                                       |
+| `LD_BUILD_EXPORT_SYMBOLS` | Whether export C++ symbols.                                                                                                                       | Off  (hidden)                                          | `LD_BUILD_SHARED_LIBS`                    |
```
kinyoklion, line 64: Hello @puncha,

First, thank you for the contribution.

The nature of this feature mean it is likely to cause crashes, memory safety issues, and other undefined behaviors. I think it is reasonable to provide this for those who understand the implications, but we will need some additional warning language.

Maybe something like `Whether to export C++ symbols in shared libraries. C++ does not have an ABI, so only use this feature if you understand the risk and requirements. A mismatch in ABI could cause crashes or other unexpected behaviors.`

AdskXiaofeng: Thank you for the review. I'll update my change soon.

puncha: @kinyoklion Sorry for the late update. Thank you for the good suggestion! I've updated the description of the cmake option. Please review again.

CMakeLists.txt
```diff
@@ -121,11 +126,14 @@ else ()
     set(Boost_USE_STATIC_LIBS ON)
 endif ()
 
-if (LD_BUILD_SHARED_LIBS)
+if (LD_BUILD_SHARED_LIBS AND (NOT LD_BUILD_EXPORT_SYMBOLS))
```
cwaldren-ld, line 129:
```suggestion
if (LD_BUILD_SHARED_LIBS AND (NOT LD_BUILD_EXPORT_CPP_SYMBOLS))
```

CMakeLists.txt
```diff
@@ -121,11 +126,14 @@ else ()
...
     # When building a shared library we hide all symbols
     # aside from this we have specifically exported for the C-API.
+    message(STATUS "LaunchDarkly: hide all symbols aside from C-API")
     set(CMAKE_CXX_VISIBILITY_PRESET hidden)
     set(CMAKE_VISIBILITY_INLINES_HIDDEN 1)
+else ()
+    message(STATUS "LaunchDarkly: expose all symbols")
 endif ()
```
cwaldren-ld, line 137:
```suggestion
if (LD_BUILD_SHARED_LIBS)
    if (LD_BUILD_EXPORT_ALL_SYMBOLS)
       message(STATUS "LaunchDarkly: exposing all symbols in shared libraries")
    else ()
       message(STATUS "LaunchDarkly: hiding all symbols in shared libraries except for C API")
       set(CMAKE_CXX_VISIBILITY_PRESET hidden)
       set(CMAKE_VISIBILITY_INLINES_HIDDEN 1)
    endif()
endif()
```

README.md
```diff
@@ -61,6 +61,7 @@ Various CMake options are available to customize the client/server SDK builds.
 | `LD_BUILD_CONTRACT_TESTS` | Whether the contract test service (used in CI) is built.                                                                                          | Off                                                    | `BUILD_TESTING`                           |
 | `LD_BUILD_EXAMPLES`       | Whether example apps (hello world) are built.                                                                                                     | On                                                     | N/A                                       |
 | `LD_BUILD_SHARED_LIBS`    | Whether the SDKs are built as static or shared libraries.                                                                                         | Off  (static lib)                                      | N/A                                       |
+| `LD_BUILD_EXPORT_SYMBOLS` | Whether to export C++ symbols in shared libraries. C++ does not have an ABI, so only use this feature if you understand the risk and requirements. A mismatch in ABI could cause crashes or other unexpected behaviors.| Off  (hidden) | `LD_BUILD_SHARED_LIBS` |
```
cwaldren-ld, line 64:
```suggestion
| `LD_BUILD_EXPORT_ALL_SYMBOLS` | Whether to export all symbols in shared libraries. By default, only C API symbols are exported because C++ does not have an ABI. Only use this feature if you understand the risk and requirements. A mismatch in ABI could cause crashes or other unexpected behaviors.| Off  (hidden) | `LD_BUILD_SHARED_LIBS` |
```

cwaldren-ld (approved):

PR #427, opened 2024-07-24
fix: Handle exception closing stream for unavailable adapter on windows.

kinyoklion:
On windows shutting down a stream for an adapter that is no longer available would crash. I was testing this by disabling adapters, but my suspicion is after a computer sleeps it may take some time for the adapters to recover on windows.

### DIFF
--- a/libs/server-sent-events/src/client.cpp
+++ b/libs/server-sent-events/src/client.cpp
@@ -349,6 +349,9 @@ class FoxyClient : public Client,
                 session_->stream.plain().cancel();
             }
         } catch (boost::system::system_error const& err) {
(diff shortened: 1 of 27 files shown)


libs/server-sent-events/src/client.cpp
```diff
@@ -341,13 +341,21 @@ class FoxyClient : public Client,
         // If any backoff is taking place, cancel that as well.
         backoff_timer_.cancel();
```
kinyoklion, line 342: This function can theoretically throw as well, but I cannot find any confirmation of a condition where it actually throws.

keelerm84 (approved):

PR #433, opened 2024-08-15
fix: handle setting protected/type-enforced attributes correctly

cwaldren-ld:
This addresses a bug where you couldn't set the `name` attribute as private using the `AttributesBuilder::SetPrivate(key, value)` API.

The issue was that this API forwards to an internal `AttributesBuilder::Set(key, value, is_private)` method, which disallowed setting of "protected" attributes.

In this PR, I've re-designated attributes as:
1. "protected": they cannot be set using the Set APIs at all
    -  This includes `_meta`, `kind`, and `key`. 
3. "typed": they can be set using Set, but must have the correct type
    - This includes `anonymous` and `name`
5. All other attributes, which don't have any type checking

Within (2), the current behavior is that both of these can be marked private, which isn't really correct. In reality only `name` can  be redacted. 

So currently, you could say `SetPrivate("anonymous, true)`, and it'd set it and add it to the private attributes list. Come redaction time, it wouldn't be redacted. 

To make this totally consistent, we'd need to disallow `SetPrivate("anonymous", ..)` as well as modify the `AddPrivateAttribute` and `AddPrivateAttributes` APIs. 

Perhaps the behavior in this PR is sufficient, given some documentation that mentions that setting `anonymous` private isn't meaningful.


libs/common/include/launchdarkly/attributes_builder.hpp
kinyoklion, line 1: The various docs need updated for this.

```
     * This method cannot be used to set the key, kind, name, or anonymous
     * property of a context. The specific methods on the context builder, or
     * attributes builder, should be used.
```

There are probably some other specific methods that need doc comment updates.

kinyoklion (changes requested):

libs/common/include/launchdarkly/attributes_builder.hpp
keelerm84: @kinyoklion I had Devin take a run at updating these docs. Seem fine to me, but can you take a look as well since this is more your area?

PR #435, opened 2024-09-03
feat: server-side payload filters

cwaldren-ld:
This PR introduces Payload Filters to the C++ Server-Side SDK. The filter key can be configured individually on the Streaming or Polling data sources. 

If an invalid key is provided (empty or not matching regex), a runtime error will be logged at the time of data source construction. The SDK will then request an unfiltered environment as a failsafe. 

e2e testing:
- [x] Manual test (streaming): valid filter key
- [x] Manual test (streaming): invalid filter key (empty, malformed)
- [x] Manual test (polling): valid filter key
- [x] Manual test (polling): invalid filter key (empty, malformed)

### DIFF
--- a/contract-tests/data-model/include/data_model/data_model.hpp
+++ b/contract-tests/data-model/include/data_model/data_model.hpp
@@ -5,31 +5,41 @@
 #include <unordered_map>
 #include "nlohmann/json.hpp"
 
-namespace nlohmann {
-
-template <typename T>
-struct adl_serializer<std::optional<T>> {
-    static void to_json(json& j, std::optional<T> const& opt) {
-        if (opt == std::nullopt) {
-            j = nullptr;
-        } else {
-            j = *opt;  // this will call adl_serializer<T>::to_json which will
-            // find the free function to_json in T's namespace!
+namespace nlohmann
+{
+    template <typename T>
+    struct adl_serializer<std::optional<T>>
+    {
+        static void to_json(json& j, std::optional<T> const& opt)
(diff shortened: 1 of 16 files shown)


contract-tests/data-model/include/data_model/data_model.hpp
```diff
@@ -6,7 +6,6 @@
 #include "nlohmann/json.hpp"
 
 namespace nlohmann {
```
cwaldren-ld, line 8: Formatter inserted these newlines. The main changes are in `Config[Streaming|Polling]Params` adding the new filter key param.

keelerm84 (approved):

libs/server-sdk/src/data_systems/background_sync/detail/payload_filter_validation/payload_filter_validation.cpp
```diff
@@ -0,0 +1,23 @@
...
+#include <boost/regex.hpp>
+
+namespace launchdarkly::server_side::data_systems::detail {
+bool ValidateFilterKey(std::string const& filter_key) {
+    if (filter_key.empty()) {
+        return false;
+    }
+    try {
+        return regex_search(filter_key,
+                            boost::regex("^[a-zA-Z0-9][._\\-a-zA-Z0-9]*$"));
```
kinyoklion, line 12: I think we should try to keep these regexes the same and `\w` more concisely represents `A-Za-z0-9_`.

PR #452, opened 2024-10-07
feat: preliminary find_package support

cwaldren-ld:
This PR adds preliminary support for:
```cmake

find_package(launchdarkly REQUIRED)

target_link_libraries(app 
        PRIVATE 
        launchdarkly::{launchdarkly-cpp-server,launchdarkly-cpp-client}
)
```
It treats the C++ monorepo as the installable unit. In the very near future, this will likely be refactored so that we have two packages. If you want both on the same host, you'd need to install both (and there'd be some redundancy.)

Currently there is only one version, `0.1.0`. This isn't updated by release-please.

For these two reasons, I'm indicating that this is preliminary and subject to change form.


CMakeLists.txt
```diff
@@ -6,7 +6,7 @@ cmake_minimum_required(VERSION 3.19)
 include(CMakeDependentOption)
 
 project(
-        LaunchDarklyCPPSDKs
```
cwaldren-ld, line 9: Attempting to unify the project name with the target alias namespace (e.g. `launchdarkly::server`).

CMakeLists.txt
```diff
@@ -6,7 +6,7 @@ cmake_minimum_required(VERSION 3.19)
...
+        launchdarkly
         VERSION 0.1
```
cwaldren-ld, line 10: Note: currently the version number isn't being updated, and I don't have a great plan for how that will work. Likely I will change to exporting two different package configs rather than the monorepo itself.

.github/actions/cmake-test/action.yml
```diff
@@ -36,6 +35,15 @@ runs:
         # along in the same manner to those test projects via the command line.
         BOOST_ROOT: ${{ steps.install-boost.outputs.BOOST_ROOT }}
         OPENSSL_ROOT_DIR: ${{ steps.install-openssl.outputs.OPENSSL_ROOT_DIR }}
+        CMAKE_INSTALL_PREFIX: ../LAUNCHDARKLY_INSTALL
```
cwaldren-ld, line 38: I'm arbitrarily installing into this folder here. Later, this folder is injected into the cmake configure step so `find_package` looks there for `launchdarklyConfig.cmake`.

That's one use-case, but it would also be good to check:
1. If you don't give it an install prefix, and just run `cmake --install .`
2. Then if you invoke `find_package` without specifying a custom `CMAKE_PREFIX_PATH`
3. It will work, due to `find_package`'s lookup logic magic.

CMakeLists.txt
```diff
@@ -155,12 +161,14 @@ endif ()
 
 set(Boost_USE_MULTITHREADED ON)
 set(Boost_USE_STATIC_RUNTIME OFF)
+
 find_package(Boost 1.81 REQUIRED COMPONENTS json url coroutine)
 message(STATUS "LaunchDarkly: using Boost v${Boost_VERSION}")
 
-include(${CMAKE_FILES}/certify.cmake)
+set(FOXY_BUILD_TESTING OFF)
 add_subdirectory(vendor/foxy)
```
cwaldren-ld, line 169: I added a new `FOXY_BUILD_TESTING` to avoid the whole "get the previous value of `BUILD_TESTING`, set it to off, `add_subdir`, set the value back.." song and dance.

README.md
```diff
@@ -66,30 +66,59 @@ Various CMake options are available to customize the client/server SDK builds.
...
+
+## Building the SDK from Source
 
-Basic usage example:
+To configure the SDK's CMake project:
 
 ```bash
-mkdir -p build && cd build
-cmake -G"Unix Makefiles" ..
+# Use 'make' as the build system.
```
cwaldren-ld, line 78: Might as well advocate for "modern" cmake commands (`-B build -S .`)

cmake/googletest.cmake
```diff
@@ -13,4 +13,8 @@ FetchContent_Declare(googletest
 )
 # For Windows: Prevent overriding the parent project's compiler/linker settings
 set(gtest_force_shared_crt ON CACHE BOOL "" FORCE)
+
+# Disable installation of googletest
```
cwaldren-ld, line 17: We could guard this with some kind of config option, that maybe is tied into `LD_BUILD_UNIT_TESTS` or something. For now, don't attempt to install googletest when the SDK is installed.

cmake/launchdarklyConfig.cmake
```diff
@@ -0,0 +1,16 @@
+include(CMakeFindDependencyMacro)
+
```
cwaldren-ld, line 2: Not entirely sure how to support this properly. With find_package, the idea is that the SDK was already built. Now, you may want to decide whether to statically or dynamically link boost/openSSL.

But that depends on how you built the SDK in the first place.

libs/client-sdk/src/CMakeLists.txt
```diff
@@ -37,7 +37,6 @@ target_sources(${LIBNAME} PRIVATE
         flag_manager/context_index.cpp
         flag_manager/flag_manager.cpp
         flag_manager/flag_persistence.cpp
-        bindings/c/sdk.cpp
```
cwaldren-ld, line 40: This was just a redundant entry.

libs/client-sdk/src/CMakeLists.txt
```diff
@@ -46,22 +45,29 @@ target_link_libraries(${LIBNAME}
         PRIVATE Boost::headers Boost::json Boost::url launchdarkly::sse launchdarkly::internal foxy)
```
cwaldren-ld, line 46: The changes in most of these sub-lib `CMakeLists.txt` are:
1. Switch from `target_include_directories(${LIBNAME} PUBLIC ../include` to one that uses generator expressions, so it works at build and installation time
2. Swap out raw `lib` or `bin` paths with `GNUInstallDir` variables
3. Install the targets to our `LD_TARGETS_EXPORT_NAME`

cmake-tests/test_find_package/project/CMakeLists.txt
```diff
@@ -0,0 +1,12 @@
+cmake_minimum_required(VERSION 3.19)
+
+project(UseFindPackageTest)
+
+find_package(launchdarkly REQUIRED)
+
+add_executable(use_find_package_server main_server.cpp)
```
cwaldren-ld, line 7: You may wonder "why launchdarkly::launchdarkly-cpp-server"? Because when I originally named the targets, I was thinking "global namespace, it might clash with someone's `server` or `client` target."

Not sure how to make this match our `launchdarkly::{client,server}` aliases that are usable in the `add_subdirectory` case.

cmake/certify.cmake
```diff
@@ -1,23 +0,0 @@
-cmake_minimum_required(VERSION 3.11)
```
cwaldren-ld, line 1: Moved into `foxy`, since this is really a foxy dependency.

vendor/foxy/cmake/certify.cmake
```diff
@@ -0,0 +1,27 @@
+cmake_minimum_required(VERSION 3.11)
+
+include(FetchContent)
+
+if (${CMAKE_VERSION} VERSION_GREATER_EQUAL "3.24")
+    # Affects robustness of timestamp checking on FetchContent dependencies.
+    cmake_policy(SET CMP0135 NEW)
+endif ()
+
+
```
cwaldren-ld, line 10: TODO: This needs to be updated to the LaunchDarkly [fork](https://github.com/launchdarkly/certify/), but there are some pulls that need to be merged before it's usable.

cmake/launchdarklyConfig.cmake
kinyoklion: Yeah, they basically just need to match.

Unless we start building a bunch of permutations. I don't think that sounds amazing though. (Boost basically does this and it never seems to work when I actually use it.)

PR #454, opened 2024-10-07
ci: fix dep graph generation when manually publishing docs

cwaldren-ld:


.github/workflows/manual-publish-doc.yml
```diff
@@ -16,11 +16,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
```
cwaldren-ld, line 18: Not entirely sure why setup-node was here. I think it can be removed

kinyoklion (approved):

PR #458, opened 2024-10-17
feat: add LDValue_SerializeJSON C binding

cwaldren-ld:
This adds a new API to the `LDValue` interface for serializing to a JSON string.

Example:

```c
LDValue foo = LDValue_NewBool(true);
char *json = LDValue_SerializeJSON(foo);
// json is the string 'true'
LDMemory_FreeString(json);
LDValue_Free(foo);
```

It's easier to view this PR commit to commit. The first one was necessary in order to access the boost::json tag_invoke implementations in `common`, where `LDValue` is implemented. Luckily, I only needed to pull over the `value` headers and some utilities. 

One thing to note is that boost serializes numbers using scientific notation (e.g. `17` serializes to `1.7E1`). Kind of an odd choice, but it is to JSON spec. 

BEGIN_COMMIT_OVERRIDE
refactor: move json_errors, primitives, and value from internal to common lib
feat: add LDValue_SerializeJSON C binding
END_COMMIT_OVERRIDE


libs/internal/include/launchdarkly/serialization/value_mapping.hpp
```diff
@@ -1,6 +1,5 @@
 #pragma once
```
cwaldren-ld, line 2: This header wasn't necessary.

kinyoklion (commented):
I am concerned about the ultimate destination of the header moves. Specifically I am concerned that it may make boost a part of our public API. I am down with providing serialization methods, but not with exposing the serialization implementation.

cwaldren-ld:
@kinyoklion, I've moved the serialization headers to `/detail/serialization` in the common lib.

kinyoklion (approved):

PR #463, opened 2024-10-25
feat: expose prerequisite relations in AllFlags API

cwaldren-ld:
Updates the `AllFlags()` API to gather and expose prerequisite information for flags. 

The prereqs are available via `State::Prerequisites()`, and are more importantly serialized in the JSON representation of the bootstrap payload. 

The implementation strategy was to create a custom `IEventProcessor` specifically for `AllFlags` usage. Previously, `AllFlags` invoked the evaluator with a no-op `EventScope`, which made event creation a no-op within the eval algorithm.

This change now passes in a `PrereqEventRecorder` which tracks the top-level prerequisites of each flag.

### DIFF
--- a/libs/server-sdk/tests/prereq_event_recorder_test.cpp
+++ b/libs/server-sdk/tests/prereq_event_recorder_test.cpp
@@ -1,0 +1,94 @@
+#include <gtest/gtest.h>
+
+#include "events/event_factory.hpp"
+#include "prereq_event_recorder/prereq_event_recorder.hpp"
+
+#include <launchdarkly/context_builder.hpp>
+
+using namespace launchdarkly;
+using namespace launchdarkly::server_side;
+using namespace launchdarkly::events;
+
+TEST(PrereqEventRecorderTest, EmptyByDefault) {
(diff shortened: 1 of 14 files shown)


libs/server-sdk/tests/prereq_event_recorder_test.cpp
```diff
@@ -0,0 +1,94 @@
...
+TEST(PrereqEventRecorderTest, IgnoresEvalEventsWithoutPrereqOf) {
+    PrereqEventRecorder recorder{"toplevel"};
+
+    auto factory = EventFactory::WithoutReasons();
+
+    auto const context = ContextBuilder().Kind("cat", "shadow").Build();
+
+    // Receiving an eval event without a prereq_of field shouldn't actually
+    // happen when calling AllFlags, but regardless we should ignore it that
+    // would signify that this isn't a prerequisite.
```
cwaldren-ld, line 65:
```suggestion
    // happen when calling AllFlags, but regardless we should ignore it because
    // that would signify that it isn't a prerequisite.
```

kinyoklion (approved):

PR #471, opened 2024-12-19
fix: Remove vendored boost file to compile with boost 1.87.

kinyoklion:
BEGIN_COMMIT_OVERRIDE
fix: Remove vendored boost file to compile with boost 1.87.
fix: Update usage of the steady_timer to be compatible with boost 1.87.
fix: Remove type alias import for JSON system_error that was removed in boost 1.87.
END_COMMIT_OVERRIDE


contract-tests/sse-contract-tests/src/event_outbox.cpp
```diff
@@ -37,13 +37,13 @@ void EventOutbox::do_shutdown(beast::error_code ec) {
 void EventOutbox::post_event(launchdarkly::sse::Event event) {
     auto http_request = build_request(callback_counter_++, std::move(event));
     outbox_.push(http_request);
-    flush_timer_.expires_from_now(boost::posix_time::milliseconds(0));
```
kinyoklion, line 40: API has been deprecated. The new API already exists in the oldest boost we support.

libs/client-sdk/src/flag_manager/flag_store.cpp
```diff
@@ -1,5 +1,3 @@
-#include <boost/json/system_error.hpp>
```
kinyoklion, line 1: This import was not used, and has been removed.

vendor/foxy/include/boost/smart_ptr/allocate_unique.hpp
kinyoklion, line 1: Appears to have been added to boost in 1.72, and the library at the time seems to have built with 1.71. So likely this was to allow working with boost 1.71 and still using this feature.

keelerm84 (approved):

PR #519, opened 2026-04-03
refactor: implement fdv2 polling initializer / synchronizer

beekld:
This implements the C++ polling initializer and synchronizer.

I have rewritten this code substantially since the first round of reviews, mostly by hand.

Changes:
* Implements `FDv2PollingInitializer` and `FDv2PollingSynchronizer`.
* Updates the `IFDv2Initializer` and `IFDv2Synchronizer` interfaces to be async.
* Makes it safe to `delete` both classes without joining any threads.
* Updates `AsioRequester::Request` to use a more modern style that is more flexible in how it can be used.
  * This will be needed for cancellation eventually anyway.
* Marks a method in `AsioRequester` as `const` to make it more obvious that it's thread-safe.
* Adds a helper to create promisified ASIO timers.
* Adds some minor Promise helpers to simplify the new code.

Caveats:
* I have not yet implemented any kind of cancellation on async operations. We should at least do this for timers at some point, but I'd rather defer it to a later PR. It's no longer a deadlock or correctness issue, but we _will_ get timeout timers slowly piling up in the background if requests are fast.

Previous description:

> The bulk of this code was generated by Claude based on the Java version, but I've gone through it line-by-line, and I think it makes sense. But I'm new to both FDv2 and ASIO, so I could be missing something.
> 
> Probably the most controversial part is the decision from the previous PR to use `std::future` and a blocking call. If we decide we need Java-like conditions, or if we need the callers to be non-blocking, we could change these to use `asio::deferred` instead. But I don't think these changes require that yet.
> 

---

> [!NOTE]
> **Medium Risk**
> Introduces new FDv2 polling data path and converts FDv2 source interfaces to `Future`-based async, which can affect orchestration flow and lifecycle/shutdown behavior. Also changes the Boost.Asio HTTP requester initiation style, so regressions could impact networking behavior and timeouts.
> 
> **Overview**
> Adds an **FDv2 polling implementation**: `FDv2PollingInitializer` performs a one-shot poll, and `FDv2PollingSynchronizer` repeatedly polls with minimum-interval enforcement and timeout/close handling using the internal `Promise/Future` utilities.
> 
> Introduces a shared `FDv2ProtocolHandler` state machine that accumulates FDv2 wire events into a complete `FDv2ChangeSet` (or emits typed protocol/JSON/server errors and `Goodbye`), with accompanying unit tests.
> 
> Refactors FDv2 source interfaces (`IFDv2Initializer::Run`, `IFDv2Synchronizer::Next`) to return `async::Future<FDv2SourceResult>` instead of blocking results, adds async helpers (`kInlineExecutor`, `MakeFuture`, `async::Delay`), and updates `AsioRequester::Request`/`Requester::Request` to be `const` and use `boost::asio::async_initiate`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4b02b670c0802f3477da78e4d9d0ef299f16ccfd. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/libs/internal/include/launchdarkly/network/asio_requester.hpp
+++ b/libs/internal/include/launchdarkly/network/asio_requester.hpp
@@ -285,11 +285,12 @@ class AsioRequester {
     auto Request(HttpRequest request, CompletionToken&& token) {
         return boost::asio::async_initiate<CompletionToken, void(HttpResult)>(
             [this](auto handler, HttpRequest req) {
-                InnerRequest(net::make_strand(ctx_), std::move(req),
-                             [h = std::move(handler)](HttpResult result) mutable {
-                                 std::move(h)(std::move(result));
-                             },
-                             0);
+                InnerRequest(
+                    net::make_strand(ctx_), std::move(req),
+                    [h = std::move(handler)](HttpResult result) mutable {
+                        std::move(h)(std::move(result));
+                    },
+                    0);
             },
             token, std::move(request));
     }
@@ -329,7 +330,7 @@ class AsioRequester {
                                  redirect_count]() mutable {
             auto beast_request = MakeBeastRequest(*request);
 
-            const auto& properties = request->Properties();
+            auto const& properties = request->Properties();
 
             std::string service =
                 request->Port().value_or(request->Https() ? "https" : "http");
--- a/libs/internal/include/launchdarkly/network/asio_requester.hpp
+++ b/libs/internal/include/launchdarkly/network/asio_requester.hpp
@@ -282,7 +282,7 @@ class AsioRequester {
     }
 
     template <typename CompletionToken>
-    auto Request(HttpRequest request, CompletionToken&& token) {
+    auto Request(HttpRequest request, CompletionToken&& token) const {
         return boost::asio::async_initiate<CompletionToken, void(HttpResult)>(
             [this](auto handler, HttpRequest req) {
                 InnerRequest(
@@ -307,7 +307,7 @@ class AsioRequester {
     void InnerRequest(boost::asio::any_io_executor exec,
                       std::optional<HttpRequest> request,
                       std::function<void(HttpResult)> callback,
-                      unsigned char redirect_count) {
+                      unsigned char redirect_count) const {
         if (redirect_count > kRedirectLimit) {
             boost::asio::post(exec, [callback, request]() mutable {
                 callback(
--- a/libs/internal/include/launchdarkly/network/asio_requester.hpp
+++ b/libs/internal/include/launchdarkly/network/asio_requester.hpp
@@ -283,22 +283,15 @@ class AsioRequester {
 
     template <typename CompletionToken>
     auto Request(HttpRequest request, CompletionToken&& token) {
-        // TODO: Clang-tidy wants to pass the request by reference, but I am not
-        // confident that lifetime would make sense.
-
-        namespace asio = boost::asio;
-        namespace system = boost::system;
-
-        using Sig = void(HttpResult result);
-        using Result = asio::async_result<std::decay_t<CompletionToken>, Sig>;
-        using Handler = typename Result::completion_handler_type;
-
-        Handler handler(std::forward<decltype(token)>(token));
-        Result result(handler);
-
-        InnerRequest(net::make_strand(ctx_), request, std::move(handler), 0);
-
-        return result.get();
+        return boost::asio::async_initiate<CompletionToken, void(HttpResult)>(
+            [this](auto handler, HttpRequest req) {
+                InnerRequest(net::make_strand(ctx_), std::move(req),
+                             [h = std::move(handler)](HttpResult result) mutable {
+                                 std::move(h)(std::move(result));
+                             },
+                             0);
+            },
+            token, std::move(request));
     }
 
    private:
--- a/libs/internal/src/fdv2_protocol_handler.cpp
+++ b/libs/internal/src/fdv2_protocol_handler.cpp
@@ -120,6 +120,10 @@ FDv2ProtocolHandler::Result FDv2ProtocolHandler::HandleServerIntent(
     }
     auto const& intent = **result;
     if (intent.payloads.empty()) {
+        // The protocol requires exactly one payload per server-intent, so
+        // an empty payloads array is a spec violation. Reset to avoid
+        // leaking accumulated state from a prior incomplete transfer.
+        Reset();
         return std::monostate{};
     }
     // The protocol defines exactly one payload per intent.
@@ -242,6 +246,8 @@ FDv2ProtocolHandler::Result FDv2ProtocolHandler::HandleGoodbye(
     auto result =
         boost::json::value_to<tl::expected<std::optional<Goodbye>, JsonError>>(
             data);
+    // Parse failures are intentionally ignored: the caller should rotate
+    // sources regardless of whether the reason field is readable.
     if (!result) {
         return Goodbye{std::nullopt};
     }
--- a/libs/internal/src/fdv2_protocol_handler.cpp
+++ b/libs/internal/src/fdv2_protocol_handler.cpp
@@ -1,0 +1,220 @@
+#include <launchdarkly/fdv2_protocol_handler.hpp>
+
+#include <launchdarkly/data_model/flag.hpp>
+#include <launchdarkly/data_model/item_descriptor.hpp>
+#include <launchdarkly/data_model/segment.hpp>
+#include <launchdarkly/serialization/json_flag.hpp>
+#include <launchdarkly/serialization/json_segment.hpp>
+
+#include <boost/json.hpp>
+#include <tl/expected.hpp>
+
+namespace launchdarkly {
+
+static char const* const kServerIntent = "server-intent";
+static char const* const kPutObject = "put-object";
+static char const* const kDeleteObject = "delete-object";
+static char const* const kPayloadTransferred = "payload-transferred";
+static char const* const kError = "error";
+static char const* const kGoodbye = "goodbye";
+
+// Returns the parsed FDv2Change on success, nullopt for unknown kinds (which
+// should be silently skipped for forward-compatibility), or an error string if
+// a known kind fails to deserialize.
+static tl::expected<std::optional<data_model::FDv2Change>, std::string>
+ParsePut(PutObject const& put) {
+    if (put.kind == "flag") {
+        auto result = boost::json::value_to<
+            tl::expected<std::optional<data_model::Flag>, JsonError>>(
+            put.object);
+        // One bad flag aborts the entire transfer so the store is never
+        // left in a partially-updated state.
+        if (!result) {
+            return tl::make_unexpected("could not deserialize flag '" +
+                                       put.key + "'");
+        }
+        if (!result->has_value()) {
+            return tl::make_unexpected("flag '" + put.key + "' object was null");
+        }
+        return data_model::FDv2Change{
+            put.key,
+            data_model::ItemDescriptor<data_model::Flag>{std::move(**result)}};
+    }
+    if (put.kind == "segment") {
+        auto result = boost::json::value_to<
+            tl::expected<std::optional<data_model::Segment>, JsonError>>(
+            put.object);
+        // One bad segment aborts the entire transfer so the store is never
+        // left in a partially-updated state.
+        if (!result) {
+            return tl::make_unexpected("could not deserialize segment '" +
+                                       put.key + "'");
+        }
+        if (!result->has_value()) {
+            return tl::make_unexpected("segment '" + put.key +
+                                       "' object was null");
+        }
+        return data_model::FDv2Change{
+            put.key,
+            data_model::ItemDescriptor<data_model::Segment>{
+                std::move(**result)}};
+    }
+    // Silently skip unknown kinds for forward-compatibility.
+    return std::nullopt;
+}
+
+static data_model::FDv2Change MakeDeleteChange(DeleteObject const& del) {
+    if (del.kind == "flag") {
+        return data_model::FDv2Change{
+            del.key,
+            data_model::ItemDescriptor<data_model::Flag>{
+                data_model::Tombstone{static_cast<uint64_t>(del.version)}}};
+    }
+    return data_model::FDv2Change{
+        del.key,
+        data_model::ItemDescriptor<data_model::Segment>{
+            data_model::Tombstone{static_cast<uint64_t>(del.version)}}};
+}
+
+FDv2ProtocolHandler::Result FDv2ProtocolHandler::HandleEvent(
+    std::string_view event_type,
+    boost::json::value const& data) {
+    if (event_type == kServerIntent) {
+        auto result = boost::json::value_to<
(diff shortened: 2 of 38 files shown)


libs/server-sdk/src/data_systems/fdv2/polling_initializer.cpp
```diff
@@ -0,0 +1,254 @@
...
+    }
+
+    auto shared_result =
+        std::make_shared<std::optional<network::HttpResult>>();
+
+    requester_.Request(request_, [this, shared_result](network::HttpResult res) {
+        std::lock_guard guard(mutex_);
+        *shared_result = std::move(res);
+        cv_.notify_one();
+    });
```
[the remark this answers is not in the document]
beekld: I want to spend a little time thinking about the best way to handle this.

libs/internal/src/fdv2_protocol_handler.cpp
```diff
@@ -0,0 +1,220 @@
...
+            return tl::make_unexpected("segment '" + put.key +
+                                       "' object was null");
+        }
+        return data_model::FDv2Change{
+            put.key,
+            data_model::ItemDescriptor<data_model::Segment>{
+                std::move(**result)}};
+    }
+    // Silently skip unknown kinds for forward-compatibility.
+    return std::nullopt;
```
kinyoklion, line 67: This ends up with maybe a slightly different layer responsibility, but I think it is likely fine. I think Java returns it, and then it is discarded a layer up, but we aren't acting on it, so it doesn't really make a difference.

kinyoklion: Actually I am a bit concerned that we may be losing granularity of data source reporting. Though maybe we don't have that concern at the moment.

libs/internal/src/fdv2_protocol_handler.cpp
```diff
@@ -0,0 +1,220 @@
...
+    return data_model::FDv2Change{
+        del.key,
+        data_model::ItemDescriptor<data_model::Segment>{
+            data_model::Tombstone{static_cast<uint64_t>(del.version)}}};
+}
+
+FDv2ProtocolHandler::Result FDv2ProtocolHandler::HandleEvent(
+    std::string_view event_type,
+    boost::json::value const& data) {
+    if (event_type == kServerIntent) {
```
kinyoklion, line 86: It might be nice to have this handle which event type it is in a more compact way, like a switch, and then handle the message is functions. Just makes it easier to see at a glance what is handled independent of how it is handled.

libs/internal/src/fdv2_protocol_handler.cpp
```diff
@@ -0,0 +1,220 @@
...
+
+    if (event_type == kDeleteObject) {
+        if (state_ == State::kInactive) {
+            return std::monostate{};
+        }
+        auto result = boost::json::value_to<
+            tl::expected<std::optional<DeleteObject>, JsonError>>(data);
+        if (!result) {
+            Reset();
+            return FDv2Error{std::nullopt, "could not deserialize delete-object"};
```
kinyoklion: We do need differentiation for malformed data errors, as we typically will restart on those errors versus just potentially logging another error type.

libs/server-sdk/src/data_systems/fdv2/polling_synchronizer.cpp
```diff
@@ -0,0 +1,294 @@
...
+
+    return HandlePollResult(*result_);
+}
+
+void FDv2PollingSynchronizer::Close() {
+    timer_.cancel();
+    std::lock_guard lock(mutex_);
+    closed_ = true;
+    cv_.notify_one();
+}
```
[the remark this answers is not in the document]
beekld: This was a real issue. I spent a bunch of time getting familiar with ASIO details and rewrote this code so that it shouldn't be a problem anymore.

libs/server-sdk/src/data_systems/fdv2/polling_initializer.cpp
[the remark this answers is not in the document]
beekld: This was intentional design, and I added comments clarifying that it is not safe to destroy this object until the ASIO thread is joined. If it turns out we need it to be destroyable earlier, we can change that later, but it adds a lot of complexity and makes `Close` slower.

libs/internal/src/fdv2_protocol_handler.cpp
```diff
@@ -0,0 +1,220 @@
...
+                        ? data_model::FDv2ChangeSet::Type::kPartial
+                        : data_model::FDv2ChangeSet::Type::kFull;
+        data_model::FDv2ChangeSet changeset{
+            type,
+            std::move(changes_),
+            data_model::Selector{data_model::Selector::State{
+                transferred.version, transferred.state}}};
+        Reset();
+        return changeset;
+    }
```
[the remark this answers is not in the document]
beekld: Fixed.

libs/server-sdk/src/data_systems/fdv2/polling_synchronizer.cpp
```diff
@@ -0,0 +1,353 @@
...
+                                boost::system::error_code) mutable {
+                    if (order[0] == 0) {
+                        promise->set_value(HandlePollResult(poll_result));
+                    } else if (closed_) {
+                        promise->set_value(
+                            FDv2SourceResult{FDv2SourceResult::Shutdown{}});
+                    } else {
+                        promise->set_value(
+                            FDv2SourceResult{FDv2SourceResult::Timeout{}});
+                    }
```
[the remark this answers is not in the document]
beekld: This is inherently a race, so I'm not sure either way is more "correct", but this is fair enough. Changed.

libs/server-sdk/src/data_systems/fdv2/polling_initializer.cpp
```diff
@@ -0,0 +1,238 @@
...
+    data_model::Selector const& selector,
+    std::optional<std::string> const& filter_key) {
+    auto url = std::make_optional(endpoints.PollingBaseUrl());
+    url = network::AppendUrl(url, kFDv2PollPath);
+
+    bool has_query = false;
+    if (selector.value) {
+        url->append("?basis=" + selector.value->state);
+        has_query = true;
+    }
```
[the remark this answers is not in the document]
beekld: Fixed.

libs/internal/src/fdv2_protocol_handler.cpp
beekld: I refactored error handling to keep a type and more context. I think that should give us what we need higher up.

libs/internal/src/fdv2_protocol_handler.cpp
beekld: Good idea. This function is way too long. I refactored it.

libs/internal/src/fdv2_protocol_handler.cpp
beekld: I'm not sure what exactly you're asking here. The Java version seems to have a distinction between parsing and "translation" that isn't there in C++. This function doesn't return raw JSON, so there's nothing to be returned in this case.

I believe, technically, our spec says that this case should silently ignore the uknown kind, whereas the Java implementation logs a warning. I followed the spec, but I could add a warning here, if you'd prefer.

This is the only branch that returns `std::nullopt`, so callers could still technically distinguish this case, but I'm not sure what else we would do about it.

libs/server-sdk/src/data_systems/fdv2/polling_synchronizer.cpp
```diff
@@ -0,0 +1,202 @@
...
+                    if (closed_) {
+                        promise->set_value(
+                            FDv2SourceResult{FDv2SourceResult::Shutdown{}});
+                    } else if (order[0] == 0) {
+                        promise->set_value(HandlePollResult(poll_result));
+                    } else {
+                        promise->set_value(
+                            FDv2SourceResult{FDv2SourceResult::Timeout{}});
+                    }
+                }));
```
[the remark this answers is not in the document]
beekld: omg, ASIO, why are you like this? 😭 

Let me consider this.

libs/server-sdk/src/data_systems/fdv2/fdv2_polling_impl.cpp
```diff
@@ -0,0 +1,194 @@
...
+    config::built::HttpProperties const& http_properties,
+    data_model::Selector const& selector,
+    std::optional<std::string> const& filter_key) {
+    auto url = std::make_optional(endpoints.PollingBaseUrl());
+    url = network::AppendUrl(url, kFDv2PollPath);
+
+    bool has_query = false;
+    if (selector.value && url) {
+        url->append("?basis=" + selector.value->state);
+        has_query = true;
```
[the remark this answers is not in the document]
beekld: Fixed.

libs/server-sdk/src/data_systems/fdv2/polling_synchronizer.cpp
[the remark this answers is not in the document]
beekld: This is a bizarre design choice by ASIO, poorly documented, and really hard to fix.

AFAICT, this is true. `wait_for_one` means that when one of the tasks completes, the others are sent a cancellation signal, but the callback isn't actually called until they all stop. But it's not trivial to handle cancellation for the http request, because of the way we use Foxy http. I think we would have to add a `Cancel` method to `FoxyClient` and then hook up all the plumbing to make it get called on cancellation.

And I don't see any other reasonable way to race futures in C++17. I guess I could create a wrapper around a `std::promise` that enforces that only the first caller gets to complete it? Because I can't block on multiple futures without occupying more threads. `std::future::then` is only available in experimental concurrency extensions.

So tomorrow, I will try two things:
1. See if Foxy supports timeouts directly, and just use that instead of treating it as a concurrency issue, or
2. Implement a `std::promise` wrapper that lets two tasks race to complete it.

libs/server-sdk/src/data_systems/fdv2/polling_initializer.hpp
```diff
@@ -0,0 +1,78 @@
...
+    };
+
+    /** Interprets an HTTP response as a source result. */
+    static data_interfaces::FDv2SourceResult HandlePollResult(
+        std::shared_ptr<State> state,
+        network::HttpResult const& res);
+
+    // Immutable state
+    network::HttpRequest const request_;
+    network::AsioRequester const requester_;
```
kinyoklion: It would be ideal if we could use `network::Requester` here instead, which supports both available networking back-ends. It does look like FDv1 polling was never moved over for server-side. The feature was mostly made for client-side, but it would be good for server-side in case anyone needs the more advanced options provided by the CURL implementation.

Could be another PR if there is a problem, but they should be reasonably interchangeable.

libs/internal/include/launchdarkly/network/asio_requester.hpp
```diff
@@ -282,23 +282,17 @@ class AsioRequester {
...
-        using Result = asio::async_result<std::decay_t<CompletionToken>, Sig>;
-        using Handler = typename Result::completion_handler_type;
-
-        Handler handler(std::forward<decltype(token)>(token));
-        Result result(handler);
-
-        InnerRequest(net::make_strand(ctx_), request, std::move(handler), 0);
-
-        return result.get();
+    auto Request(HttpRequest request, CompletionToken&& token) const {
```
kinyoklion, line 285: We will want to be sure to keep these in lock-step with the curl, and generic wrapper versions of request handling.

libs/server-sdk/src/data_systems/fdv2/polling_initializer.hpp
beekld: Excellent point. Done.

libs/internal/include/launchdarkly/network/asio_requester.hpp
beekld: Fortunately, the curl implementation was already marked as const, so I moved the const up to the shared interface. It's nice because it makes it easier to assume thread-safety at a glace.

(The other changes in that file are from clang auto-formatting.)

libs/internal/src/fdv2_protocol_handler.cpp
kinyoklion: So, the thing I am trying to get at here is that the protocol handler here knows about the types, where I expect 1 layer up from here to know the types. Or for this to take a mapper type interface.

The reason being is that theoretically a client and server can use the same protocol handler, but they get different model types.

Server has `flag` and `segment`, and client-side has a `flag-eval`. The protocol itself should be agnostic to these types theoretically. Then you report "hey we got a change to kind: potato", and the data source layer can be like "I don't support potato" and then discard it. 

In the Java version it just remains as some kind of JsonElement until another layer knows how to handle it.

Another option though is a mapper. Where the handler takes a mapper, and then says "map this type for me" and if the mapper says it cannot handle it, then we also know to discard it.

Mapper approach:
```
  const protocolHandler = internal.createProtocolHandler(
    { 'flag-eval': processFlagEval },
    config.logger,
  );
```

In regards to the log, I do tend to prefer a log, but one could argue about it being debug, or just once, or something. But that isn't the main concern. This layer knowing about the specific types is.

libs/server-sdk/src/data_systems/fdv2/fdv2_polling_impl.cpp
```diff
@@ -0,0 +1,198 @@
...
+        auto const& body = res.Body();
+        if (!body) {
+            return FDv2SourceResult{FDv2SourceResult::Interrupted{
+                MakeError(ErrorKind::kInvalidData, 0,
+                          "polling response contained no body"),
+                false}};
+        }
+
+        boost::system::error_code ec;
+        auto parsed = boost::json::parse(*body, ec);
```
kinyoklion: Maybe we could lift the parsing into a function or have its own type?

kinyoklion (approved):

libs/internal/src/fdv2_protocol_handler.cpp
```diff
@@ -0,0 +1,265 @@
...
+        return data_model::FDv2Change{
+            del.key,
+            data_model::ItemDescriptor<data_model::Flag>{
+                data_model::Tombstone{static_cast<uint64_t>(del.version)}}};
+    }
+    return data_model::FDv2Change{
+        del.key,
+        data_model::ItemDescriptor<data_model::Segment>{
+            data_model::Tombstone{static_cast<uint64_t>(del.version)}}};
+}
```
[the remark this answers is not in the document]
beekld: This was intentional, because the only other reasonable thing to do here is assert, and the types are checked before we get here. I just don't think this is worth changing.

libs/server-sdk/src/data_systems/fdv2/fdv2_polling_impl.cpp
beekld: Done.

libs/internal/src/fdv2_protocol_handler.cpp
beekld: I am making a note to handle this in a follow-up PR. I'm just worried about this PR getting too big and out of hand.

PR #521, opened 2026-04-13
refactor: add internal Promise/Future classes for future use

beekld:
This PR adds internal-only `Promise` and `Future` classes. These have some advantages over using callbacks everywhere:
* They are more easily composable. That means utility functions like `WhenAll` and `WhenAny` only have to be written once, instead of having to re-invent the wheel every time we need those concepts.
* This allows C++ code to be structured more like the code in other languages with proper promise types.
* Unlike `std::future`, this supports a `Then` method, which enables things like `WhenAny` without occupying an extra thread per future.
* Unlike `std::function`, the `Continuation` class used in this implementation supports move-only functions.

Some points to consider:
* Would other people like using these, or am I the only one?
* **Naming**: I chose to mimic the C++ names of `std::promise`, `std::future`, and `then()` (experimental).
* **Executors**: I chose to _require_ passing an executor into `Then`, because there is no natural default:
  * C++ doesn't have a "main" thread exactly, or at least, it's not used the same way as on other platforms.
  * Immediate executors are confusing to reason about, and result in calling work on unpredictable threads.
  * We shouldn't assume `asio::post`, as this class may be used for other things.
  * I didn't want to add the complexity of storing execution contexts in a thread-local.
* **Error handling**: I chose not to put error handling directly in the `Future`, and encourage users to use `Future<std::expected<>>` if they care about it, to keep things simpler.
* **Cancellation**: I chose not to implement cancellation. We can add it in the future. But cancellation can be treated as mostly orthogonal to futures, and we can't cancel http requests currently anyway, so 🤷‍♀️ .
* By default, `GetResult` is non-blocking, as this prevents unintended blocking. There is a blocking version called `WaitForResult` with a timeout if needed.
* The continuations in `Then` are passed the _result_ of the `Future` instead of the `Future` itself. This is simpler, but it means we can't add more metadata in the future without a breaking change. This is an internal class, so I think it's better to keep it simple.
* The type `T` in `Promise<T>`/`Future<T>` has to be copyable. This is an inherent limitation of the need to be able to have `Future<Future<T>>` unwrappable to `Future<T>`, while also letting a `Future` have multiple continuations.
* The executor has to be copyable, for simplicity. ASIO has the same requirement on executors, and it seems fine.

I added lots of tests. Some of them double as documentation for humans and LLMs.

---

> [!NOTE]
> **Medium Risk**
> Introduces new concurrency primitives (`Promise`/`Future`, move-only `Continuation`, and `WhenAll`/`WhenAny`) with mutex/atomic coordination and executor-dispatched continuations, so correctness depends on subtle threading and lifetime behavior. Changes are currently additive/internal but could impact future async flows once adopted.
> 
> **Overview**
> Adds a new internal async module `launchdarkly/async/promise.hpp` implementing `Promise<T>`/`Future<T>` with executor-based `Then` chaining (including flattening when a continuation returns another `Future`) and a move-only `Continuation` wrapper to support move-only captures.
> 
> Introduces aggregation helpers `WhenAll` and `WhenAny`, a blocking `WaitForResult` (timeout-based) for test/support scenarios, and wires the new async headers into the internal library build. Adds extensive unit tests covering chaining, move/copy behavior, concurrent resolution, and ASIO executor usage.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0dfce8854c628cf06c17e6f5731359abadc777a8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/libs/internal/include/launchdarkly/async/promise.hpp
+++ b/libs/internal/include/launchdarkly/async/promise.hpp
@@ -13,6 +13,14 @@ namespace launchdarkly::async {
 // requires all captured variables to be copy-constructible, which prevents
 // storing lambdas that capture move-only types.
 //
+// Continuation is used to represent units of work passed to executors. An
+// executor is a callable with signature void(Continuation<void()>) that
+// schedules the work to run somewhere — for example, on an ASIO io_context:
+//
+//   auto executor = [&ioc](Continuation<void()> work) {
+//       boost::asio::post(ioc, std::move(work));
+//   };
+//
 // The primary template is declared but not defined; only the partial
 // specialization below (which splits Sig into R and Args...) is usable.
 // This lets callers write Continuation<void()> instead of Continuation<void>.
@@ -39,6 +47,8 @@ class Continuation<R(Args...)> {
     std::unique_ptr<Base> impl_;
 
    public:
+    // Constructs a Continuation from any callable F. F may be a lambda,
+    // function pointer, or other callable; it need not be copy-constructible.
     // F&& is a forwarding reference: accepts any callable by move or copy,
     // then moves it into Impl<F> so Continuation itself owns the callable.
     template <typename F>
@@ -47,6 +57,8 @@ class Continuation<R(Args...)> {
     Continuation(Continuation&&) = default;
     Continuation& operator=(Continuation&&) = default;
 
+    // Invokes the stored callable with the given arguments.
+    // Returns whatever the callable returns.
     R operator()(Args... args) {
         return impl_->call(std::forward<Args>(args)...);
     }
@@ -76,9 +88,17 @@ struct future_value<Future<T>> {
     using type = T;
 };
 
-// TODO: Document why things are what they are.
+// PromiseInternal holds the shared state between a Promise and its associated
+// Futures: the result value, the mutex protecting it, and the list of
+// continuations waiting to run when the result is set.
+//
+// This is an internal class and not intended to be used directly. Promise and
+// Future each hold a shared_ptr<PromiseInternal>, which lets multiple Future
+// copies all refer to the same underlying state, and lets Promise and Future
+// have independent lifetimes while the shared state remains alive as long as
+// either end holds it.
+//
 // TODO: Is it okay that the shared_ptr lets copies mutate the result??
-
 template <typename T>
 class PromiseInternal {
    public:
@@ -87,6 +107,9 @@ class PromiseInternal {
     PromiseInternal(PromiseInternal const&) = delete;
     PromiseInternal(PromiseInternal&&) = delete;
 
+    // Sets the result and schedules all registered continuations via their
+    // executors. Returns true if the result was set, or false if it was already
+    // set by a previous call to Resolve.
     bool Resolve(T result) {
         std::lock_guard lock(mutex_);
         if (result_->has_value()) {
@@ -103,11 +126,14 @@ class PromiseInternal {
         return true;
     }
 
+    // Returns true if Resolve has been called.
     bool IsFinished() const {
         std::lock_guard lock(mutex_);
         return result_->has_value();
     }
 
+    // Returns a reference to the result. The caller must ensure IsFinished()
+    // is true before calling this.
     T const& GetResult() const {
         std::lock_guard lock(mutex_);
         return **result_;
@@ -208,6 +234,21 @@ class PromiseInternal {
         continuations_;
 };
 
+// Promise is the write end of a one-shot async value, similar to std::promise.
+// Create a Promise<T>, hand its Future to a consumer via GetFuture(), then
+// call Resolve() exactly once to deliver the value.
+//
+// Promise is move-only: it cannot be copied, but it can be moved. This
+// prevents accidentally resolving the same promise from two places.
+//
+// Using std::expected<V, E> as T is the recommended way to represent
+// operations that may fail:
+//
+//   Promise<std::expected<int, std::string>> promise;
+//   Future<std::expected<int, std::string>> future = promise.GetFuture();
+//   // ... hand future to a consumer, then later:
+//   promise.Resolve(42);                            // success
+//   promise.Resolve(std::unexpected("timed out"));  // failure
 template <typename T>
 class Promise {
    public:
@@ -216,14 +257,44 @@ class Promise {
     Promise(Promise const&) = delete;
     Promise(Promise&&) = default;
 
+    // Sets the result to the given value and schedules any continuations that
+    // were registered via Future::Then. Returns true if the result was set, or
+    // false if Resolve was already called.
     bool Resolve(T result) { return internal_->Resolve(result); }
 
+    // Returns a Future that will resolve when this Promise is resolved.
+    // May be called multiple times; each call returns a Future referring to
+    // the same underlying state.
     Future<T> GetFuture() { return Future(internal_); }
 
    private:
     std::shared_ptr<PromiseInternal<T>> internal_;
 };
 
+// Future is the read end of a one-shot async value, similar to std::future,
+// but with support for chaining via Then.
+//
+// A Future is obtained from Promise::GetFuture(). Multiple copies of a Future
+// may exist and all refer to the same underlying result. When the associated
+// Promise is resolved, all continuations registered via Then are scheduled.
+//
+// Unlike std::future, Future does not support blocking on the result directly.
+// Instead, use Then to attach work that runs once the value is available.
+//
+// Example using std::expected<V, E> to represent a fallible async operation:
+//
+//   boost::asio::io_context ioc;
+//   auto executor = [&ioc](Continuation<void()> work) {
+//       boost::asio::post(ioc, std::move(work));
+//   };
+//
+//   Future<std::expected<float, std::string>> result = future.Then(
+//       [](std::expected<int, std::string> const& val)
+//               -> std::expected<float, std::string> {
+//           if (!val) return std::unexpected(val.error());
+//           return *val * 1.5f;
+//       },
+//       executor);
 template <typename T>
 class Future {
    public:
@@ -233,11 +304,24 @@ class Future {
     Future(Future const&) = default;
     Future(Future&&) = default;
 
+    // Returns true if the associated Promise has been resolved.
     bool IsFinished() const { return internal_->IsFinished(); }
 
+    // Returns a reference to the result. The caller must ensure IsFinished()
+    // is true before calling this. The reference is valid for the lifetime of
+    // the Future (or any copy of it).
     T const& GetResult() const { return internal_->GetResult(); }
 
-    // Then where the continuation returns R directly, yielding Future<R>.
+    // Registers a continuation to run when this Future resolves, returning a
+    // new Future<R> that resolves to the continuation's return value.
+    //
+    // Parameters:
+    //   continuation - Called with the resolved T const& when this Future
+    //                  resolves. Must return a value of type R (not a Future).
+    //   executor     - Called with the work to schedule when this Future
+    //                  resolves. Controls where and when the continuation runs.
+    //
+    // Returns a Future<R> that resolves to whatever the continuation returns.
     template <typename F,
               // Deduce R from what F returns when called with T const&.
               typename R = std::invoke_result_t<F, T const&>,
@@ -249,8 +333,25 @@ class Future {
                                std::move(executor));
     }
 
-    // Then where the continuation returns Future<T2> (i.e. R = Future<T2>),
-    // yielding a flattened Future<T2> that resolves when the inner future does.
+    // Registers a continuation to run when this Future resolves, where the
+    // continuation itself returns a Future<T2>. Returns a flattened Future<T2>
+    // that resolves when the inner future does, avoiding Future<Future<T2>>.
+    // Use this overload to chain async operations that themselves return a
+    // Future, avoiding a nested Future<Future<T2>>:
+    //
+    //   Future<std::expected<Data, Err>> result = future.Then(
+    //       [](std::expected<Key, Err> const& key) {
+    //           return fetch(key);  // fetch returns Future<expected<Data, Err>>
+    //       },
+    //       executor);
+    //
+    // Parameters:
+    //   continuation - Called with the resolved T const& when this Future
+    //                  resolves. Must return a Future<T2>.
+    //   executor     - Called with the work to schedule when this Future
+    //                  resolves, and again when the inner Future resolves.
+    //
+    // Returns a Future<T2> that resolves when the inner Future<T2> resolves.
     template <typename F,
               // Deduce R from what F returns when called with T const&.
               typename R = std::invoke_result_t<F, T const&>,
--- a/libs/internal/include/launchdarkly/async/promise.hpp
+++ b/libs/internal/include/launchdarkly/async/promise.hpp
@@ -114,22 +114,29 @@ class PromiseInternal {
     // executors. Returns true if the result was set, or false if it was already
     // set by a previous call to Resolve.
     bool Resolve(T result) {
-        std::lock_guard lock(mutex_);
-        if (result_->has_value()) {
-            return false;
+        std::vector<Continuation<void(std::shared_ptr<std::optional<T>>)>>
+            to_call;
+        {
+            std::lock_guard lock(mutex_);
+            if (result_->has_value()) {
+                return false;
+            }
+
+            // Move result into storage if possible; otherwise copy.
+            if constexpr (std::is_move_assignable_v<T>) {
+                *result_ = std::move(result);
+            } else {
+                *result_ = result;
+            }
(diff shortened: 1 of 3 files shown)


beekld:
The Windows CI seems to be broken because it couldn't download OpenSSL? I don't think that has anything to do with my change.

beekld:
FWIW, I wrote the bulk of this code by hand. Claude helped fill out the comments, and also helped me work through some of the trickier template syntax.

libs/internal/include/launchdarkly/async/promise.hpp
```diff
@@ -0,0 +1,513 @@
...
+        // Move result into storage if possible; otherwise copy.
+        if constexpr (std::is_move_assignable_v<T>) {
+            *result_ = std::move(result);
+        } else {
+            *result_ = result;
+        }
+
+        for (auto& continuation : continuations_) {
+            continuation(result_);
+        }
```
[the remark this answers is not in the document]
beekld: This is a real concern when using an immediate executor. What I don't understand yet is why the existing test didn't catch it.

Anyway, I moved the continuation calls outside of the mutex, since that's the right thing to do regardless.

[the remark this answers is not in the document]
beekld: I went back and reran those tests and they do deadlock, so it must've been user error.

libs/internal/include/launchdarkly/async/promise.hpp
```diff
@@ -0,0 +1,535 @@
...
+    std::unique_ptr<Base> impl_;
+
+   public:
+    // Constructs a Continuation from any callable F. F may be a lambda,
+    // function pointer, or other callable; it need not be copy-constructible.
+    // F&& is a forwarding reference: accepts any callable by move or copy,
+    // then moves it into Impl<F> so Continuation itself owns the callable.
+    template <typename F>
+    Continuation(F&& f)
+        : impl_(std::make_unique<Impl<F>>(std::forward<F>(f))) {}
```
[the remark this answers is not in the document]
beekld: Fixed. And test added.

libs/internal/include/launchdarkly/async/promise.hpp
```diff
@@ -0,0 +1,535 @@
...
+//   // ... hand future to a consumer, then later:
+//   promise.Resolve(42);                            // success
+//   promise.Resolve(tl::unexpected("timed out"));  // failure
+template <typename T>
+class Promise {
+   public:
+    Promise() : internal_(new PromiseInternal<T>()) {}
+    ~Promise() = default;
+    Promise(Promise const&) = delete;
+    Promise(Promise&&) = default;
```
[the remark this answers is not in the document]
beekld: How did I miss that? Fixed.

libs/internal/include/launchdarkly/async/promise.hpp
```diff
@@ -0,0 +1,535 @@
...
+        return result_->has_value();
+    }
+
+    // Returns a const reference to the optional result. The optional contains
+    // a value if resolved, or is empty if not yet resolved. The reference is
+    // valid for the lifetime of any Future referring to this PromiseInternal.
+    std::optional<T> const& GetResult() const {
+        std::lock_guard lock(mutex_);
+        return *result_;
+    }
```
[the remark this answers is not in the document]
beekld: I really wish cursor would batch up all its feedback at once. I also wish it could give me the feedback before I officially open the review.

Anyway, yes, this is a real race, but only if the Promise wasn't resolved. Fixed by returning a separate empty member if it wasn't. That fixes the race without changing the contract.

libs/internal/include/launchdarkly/async/promise.hpp
```diff
@@ -0,0 +1,548 @@
...
+
+        auto do_work = [outerPromise = std::move(outerPromise),
+                        continuation = std::move(continuation),
+                        executor](T const& val) mutable {
+            Future<T2> innerFuture = continuation(val);
+            innerFuture.Then(
+                [outerPromise = std::move(outerPromise)](
+                    T2 const& inner_val) mutable -> T2 {
+                    outerPromise.Resolve(inner_val);
+                    return inner_val;
```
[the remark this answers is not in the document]
beekld: I wish this one had been in the first round of feedback. This is really tricky, because we can't take ownership of the result underlying the future handed to us. I think this may make it impossible for futures to actually work with move-only types. I'm going to think about this some more.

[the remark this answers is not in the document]
beekld: Yeah, I'm gonna sleep on it, but I think I'm going to change this to require `T` to be copyable. That will simplify a lot of the code. But I don't see any other way to allow flattening a `Future<Future<T>>` where the `Future` has shared state with a `Promise`.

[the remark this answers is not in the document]
beekld: I gave it a lot of thought, and I think `T` has to be copyable. Done.

libs/internal/include/launchdarkly/async/promise.hpp
```diff
@@ -0,0 +1,529 @@
...
+            } else {
+                result_ = result;
+            }
+            to_call = std::move(continuations_);
+        }
+
+        // Call continuations outside the lock so that continuations which
+        // re-enter this future (e.g. via GetResult or Then) don't deadlock.
+        for (auto& continuation : to_call) {
+            continuation(*result_);
```
kinyoklion, line 138: Should we copy the result to a local under the lock?

```
  bool Resolve(T result) {                                                                                     
     ...
      T resolved_value;  // local copy                                                                         
      {
          ...                                                               
          resolved_value = *result_;  // copy under the lock
          ...                                                          
      }
      for (auto& continuation : to_call) {                                                                     
          continuation(resolved_value);  // use local copy                                                     
      }
      return true;                                                                                             
  }   
```

libs/internal/include/launchdarkly/async/promise.hpp
```diff
@@ -0,0 +1,529 @@
...
+            std::mutex mutex;
+            std::condition_variable cv;
+            bool ready = false;
+        };
+        auto state = std::make_shared<State>();
+
+        // The continuation ignores T entirely (returning a dummy int). The
+        // executor signals the cv when called, since being called means the
+        // original future has resolved.
+        Then([](T const&) { return 0; },
```
kinyoklion: Would it make sense to have something similar to an `OnResolved` that doesn't need to create the inner Promise<int>/Future<int>?

Probably not I guess.

libs/internal/include/launchdarkly/async/promise.hpp
```diff
@@ -0,0 +1,529 @@
...
+//       [](tl::expected<int, std::string> const& val)
+//               -> tl::expected<float, std::string> {
+//           if (!val) return tl::unexpected(val.error());
+//           return *val * 1.5f;
+//       },
+//       executor);
+template <typename T>
+class Future {
+   public:
+    Future(std::shared_ptr<PromiseInternal<T>> internal)
```
kinyoklion, line 335: Should we be doing the fiend-class dance to enforce use of ::GetFuture? Internal, so probably not critical.

libs/internal/include/launchdarkly/async/promise.hpp
```diff
@@ -0,0 +1,529 @@
...
+//   Promise<tl::expected<int, std::string>> promise;
+//   Future<tl::expected<int, std::string>> future = promise.GetFuture();
+//   // ... hand future to a consumer, then later:
+//   promise.Resolve(42);                            // success
+//   promise.Resolve(tl::unexpected("timed out"));  // failure
+template <typename T>
+class Promise {
+   public:
+    Promise() : internal_(std::make_shared<PromiseInternal<T>>()) {}
+    ~Promise() = default;
```
kinyoklion, line 282: Do we need any Future handling for an abandoned unresolved promise?

libs/internal/include/launchdarkly/async/promise.hpp
```diff
@@ -0,0 +1,529 @@
...
+    bool Resolve(T result) {
+        std::vector<Continuation<void(T const&)>> to_call;
+        {
+            std::lock_guard lock(mutex_);
+            if (result_.has_value()) {
+                return false;
+            }
+
+            // Move result into storage if possible; otherwise copy.
+            if constexpr (std::is_move_assignable_v<T>) {
```
kinyoklion: Do we need this check?

libs/internal/include/launchdarkly/async/promise.hpp
```diff
@@ -0,0 +1,529 @@
...
+                    std::function<void(Continuation<void()>)> executor) {
+        return internal_->Then(std::forward<F>(continuation),
+                               std::move(executor));
+    }
+
+   private:
+    std::shared_ptr<PromiseInternal<T>> internal_;
+};
+
+// WhenAll takes a variadic list of Futures (each with potentially different
```
kinyoklion, line 440: Should we clarify the executor here?

libs/internal/tests/promise_test.cpp
kinyoklion, line 1: Do we need some more thread-intense test scenarios?

libs/internal/include/launchdarkly/async/promise.hpp
beekld: This was intentional, to avoid an unnecessary copy. It's safe to access `result_` outside the lock here because once `result_` is set (which is checked above in the lock), it can never be changed. Due to the nature of promises, I can't imagine that invariant ever changing.

Would you rather I add a comment to that a effect, or make the extra copy to be defensive?

libs/internal/include/launchdarkly/async/promise.hpp
beekld: I don't feel strongly either way. I would probably lean towards not adding it unless we see real demand for it.

I changed this one to use `std::monostate{}`, because I think that's a little more clearly a stand-in for `void` than using an `int`.

libs/internal/include/launchdarkly/async/promise.hpp
beekld: I don't think that's necessary, because people should already not be using `PromiseInternal` directly. I wish we could enforce that part, but I don't think there's a way to do that when it's a templated class in a header file?

libs/internal/include/launchdarkly/async/promise.hpp
beekld: This was intentional, to avoid copying the result if its a moveable value. Normally I wouldn't bother with this kind of optimization, but since it's already in a class full of generic template magic, I figured it fit in. Before changing this to require copyable types, this was necessary, but now it's just a minor optimization.

But I don't feel strongly about it, if you'd rather we just always copy.

libs/internal/include/launchdarkly/async/promise.hpp
beekld: I don't think that's necessary. The only executor used here is an immediate executor that resolves the promise whose future is returned. We really don't care which thread that code runs on. When callers use `Then` on the returned future, they can specify their own executor for that.

libs/internal/include/launchdarkly/async/promise.hpp
beekld: That's a really interesting question. If we had a standardized way to signal failure, I would try signaling all futures with that failure. If this were Java, I would fail them with a `RuntimeException`... But since the C++ standard is `std::expected<T, E>` with no constraint on `E`, I'm not really sure what we can do. I don't really think it's worth being more opinionated about error handling here.

TBH, I'm not even sure it's always a failure, since the caller could be racing multiple futures and not care if one of them was abandoned. 🤷‍♀️ 

I'm open to discussing it more, if you have ideas.

libs/internal/tests/promise_test.cpp
beekld: If you've got any suggestions, I'm happy to take them. I added a test that spawns more threads. Not really sure what other scenarios to try.

FWIW, I tried to run the tests with tsan, but this build is configured with asan, and they can't both be used together. I can look into that more, if you think it's useful. Might need to add a second build configuration or something.

libs/internal/include/launchdarkly/async/promise.hpp
kinyoklion: I am just a little concerned with what the fences may look like here, and a copy inside the lock removes that concern. But I cannot really imagine a re-ordering situation where it would be a problem.

libs/internal/include/launchdarkly/async/promise.hpp
kinyoklion: I think its only a problem if someone was waiting it? Because it would never resolve? We could handle it "application" level and just have some assert?

Like, you always have to signal, and that means that your types have to be something that can report failure?

libs/internal/include/launchdarkly/async/promise.hpp
kinyoklion: I was thinking always just std::move? Because it should be safe to attempt to move an unmovable type, because it would be equivalent to a cast to an rvalue and just copy when it wasn't movable?

libs/internal/tests/promise_test.cpp
kinyoklion: It could be useful, unfortunately we have lots of builds. But enforcing as much correctness as we can is good.

I'll look at the tests again to see if I have any better thread test suggestions.

libs/internal/include/launchdarkly/async/promise.hpp
kinyoklion: Through everything having protected/private constructors and using friends. You can see the type but not really do much with it.

But I think its is fine as is. People shouldn't be using things from internal.

libs/internal/tests/promise_test.cpp
kinyoklion: Not really threading, but do we have a test for the timeout path when waiting for a result?

I am not sure if the thread scenarios would really cover more ground. 

Creating continuations on multiple threads, racing resolutions on multiple threads, using a multi-threaded asio executor.

They are also probably situations we won't be encountering.

kinyoklion (approved):
Remaining comments are all optional.

libs/internal/include/launchdarkly/async/promise.hpp
beekld: Yeah, it's only a problem if someone is waiting, and even then `WaitForResult` has a timeout. I dunno, I just think if we want to be opinionated about error handling in this class, we should give it some thought. I'll go with this for now, and we can revisit it later if it turns out to be a problem.

FWIW, it looks like this SDK uses `expected` with both `launchdarkly::Error` and `launchdarkly::JsonError` quite a bit. If we were standardized on one of those, I'd probably be less resistant to using it directly. It might be worth standardizing on a single error enum across the codebase at some point in the future.

libs/internal/include/launchdarkly/async/promise.hpp
beekld: I tried that first, but it turns out it doesn't compile, because of the way `std::optional` is written. :-/

I actually have a test that catches it.

libs/internal/include/launchdarkly/async/promise.hpp
beekld: If the value were a known type, I'd probably just copy it to be on the safe side. But since it's not, I'll leave this for now. Happy to revisit in the future.

libs/internal/tests/promise_test.cpp
beekld: I went ahead and added some tests for those cases. Doesn't hurt to be thorough when it comes to thread safety. 🫡

libs/internal/include/launchdarkly/async/promise.hpp
```diff
@@ -0,0 +1,531 @@
...
+            if (result_.has_value()) {
+                return false;
+            }
+
+            // Move result into storage if possible; otherwise copy.
+            if constexpr (std::is_move_assignable_v<T>) {
+                result_ = std::move(result);
+            } else {
+                result_ = result;
+            }
```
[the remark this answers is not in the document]
beekld: I asked Claude about this, and it said it would be "pathological" to make a class that is move assignable but not move constructible. 😂  Nevertheless, I fixed this.

libs/internal/tests/promise_test.cpp
```diff
@@ -0,0 +1,562 @@
...
+
+// Verifies that multiple threads can concurrently register continuations on
+// the same future before it resolves, and all continuations run after
+// resolution.
+TEST(Promise, ConcurrentThenRegistration) {
+    Promise<int> promise;
+    Future<int> future = promise.GetFuture();
+
+    std::atomic<int> count{0};
+    std::vector<std::thread> threads;
```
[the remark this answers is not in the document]
beekld: Removed the test. Tried fixing it, but it's too complicated and not worth it.

PR #524, opened 2026-04-21
refactor: Implement cancellation for Delay() timers.

beekld:
This implements CancellationSource / CancellationToken, based on the similar concepts in C++20 and other languages. It also updates the spots where we race timers to cancel any timers that aren't finished, so that they don't pile up internally.

This is mostly Claude-generated code, but I have read through it and verified it is correct.

There are alternatives to this model for simpler cases, but I like that this implementation means we could eventually use it for other cases (like cancelling http requests), and since all the things racing share a cancellation source, they can all be cancelled together.

This PR does not implement cancellation for HTTP requests yet, since that is much more complicated, and because they are less likely to pile up during normal operation.

---

> [!NOTE]
> **Medium Risk**
> Introduces new cross-thread cancellation primitives and changes timer scheduling/cancellation behavior, which can affect concurrency, ordering, and potential deadlocks/leaks if incorrect. Scope is limited to async utilities and FDv2 polling timeouts/delays, with added tests to validate the new behavior.
> 
> **Overview**
> Adds a new `CancellationSource`/`CancellationToken`/`CancellationCallback` primitive (C++20 `stop_*`-style) to support coordinated cancellation with RAII callback registration and destructor synchronization.
> 
> Extends `async::Delay` to accept an optional `CancellationToken` and cancels the underlying ASIO timer when triggered, while also hardening `Continuation` construction to avoid self-recursion. Updates FDv2 polling (`polling_synchronizer`) to create a shared cancellation source for each `WhenAny` race and cancel the non-winning delay/timeout timers to prevent timer buildup, and adds comprehensive unit tests for cancellation and timer interaction.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a187254eaa28bd4a00bcf2b7dcc11811d631be05. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/libs/internal/include/launchdarkly/async/timer.hpp
+++ b/libs/internal/include/launchdarkly/async/timer.hpp
@@ -1,0 +1,32 @@
+#pragma once
+
+#include <launchdarkly/async/promise.hpp>
+
+#include <boost/asio/any_io_executor.hpp>
+#include <boost/asio/error.hpp>
+#include <boost/asio/steady_timer.hpp>
+#include <boost/system/error_code.hpp>
+
+#include <chrono>
+#include <memory>
+
+namespace launchdarkly::async {
+
+// Returns a Future<bool> that resolves once the given duration elapses.
+// The future resolves with true if the timer fired normally, or false if
+// the timer was cancelled before it expired.
+template <typename Rep, typename Period>
+Future<bool> Delay(boost::asio::any_io_executor executor,
+                   std::chrono::duration<Rep, Period> duration) {
+    auto timer = std::make_shared<boost::asio::steady_timer>(executor);
+    timer->expires_after(duration);
+    Promise<bool> promise;
+    auto future = promise.GetFuture();
+    timer->async_wait([p = std::move(promise),
+                       timer](boost::system::error_code code) mutable {
+        p.Resolve(code != boost::asio::error::operation_aborted);
+    });
+    return future;
+}
+
+}  // namespace launchdarkly::async
--- a/libs/internal/include/launchdarkly/async/timer.hpp
+++ b/libs/internal/include/launchdarkly/async/timer.hpp
@@ -1,9 +1,11 @@
 #pragma once
 
+#include <launchdarkly/async/cancellation.hpp>
 #include <launchdarkly/async/promise.hpp>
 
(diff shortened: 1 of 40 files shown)


libs/internal/include/launchdarkly/async/timer.hpp
```diff
@@ -15,17 +17,55 @@ namespace launchdarkly::async {
...
+    auto cancel_timer = [timer, executor,
+                         timer_started_future =
+                             std::move(timer_started_future)]() mutable {
+        timer_started_future.Then(
+            [timer](auto const&) -> std::monostate {
+                timer->cancel();
+                return {};
+            },
+            [executor](Continuation<void()> f) {
+                boost::asio::post(executor, f);
```
kinyoklion: Without a move here there seems to be a recursive construction of the continuation. But maybe my test scenario isn't valid?

```
TEST(ContinuationRecursion, TimerCancelTriggersStackOverflow) {
    boost::asio::io_context ioc;
    auto work = boost::asio::make_work_guard(ioc);
    std::thread t([&] { ioc.run(); });

    CancellationSource cancel;

    auto future = Delay(ioc.get_executor(), std::chrono::seconds(60),
                        cancel.GetToken());
    
    std::this_thread::sleep_for(std::chrono::milliseconds(50));
    cancel.Cancel();

    const auto result = future.WaitForResult(std::chrono::seconds(5));

    // Doesn't make it here.

    ASSERT_TRUE(result.has_value());
    EXPECT_FALSE(*result);

    work.reset();
    t.join();
}
```

Image

kinyoklion (changes requested):
Feedback in comment.

kinyoklion (approved):

libs/internal/include/launchdarkly/async/timer.hpp
beekld: Wow, good catch. That was kind of subtle.

Fixed.
* Added a template constraint so that this will give a compile error instead of a runtime error.
* Changed `f` to `std::move(f)` so it compiles.
* Added the test above, which failed before and passes after.

PR #525, opened 2026-04-22
refactor: split json parsing/translation for future-proofing client/server distinction

beekld:
As discussed in a previous PR, this splits apart the parsing and translation for FDv2 changesets, so that the server-specific code is separate from the shared code. This will be needed once we add client code later.

I went with passing around JSON objects everywhere, rather than passing in a translator, because 1) it's simpler, and 2) it's more like Java, which has advantages for understanding code across codebases.

---

> [!NOTE]
> **Medium Risk**
> Moderate risk because it changes the FDv2 ingestion pipeline and error/forward-compatibility behavior (unknown kinds and JSON deserialization), which could affect store updates and sync reliability.
> 
> **Overview**
> **Splits FDv2 event handling into two phases**: the shared `FDv2ProtocolHandler` now only accumulates raw changes (`kind`, `key`, `version`, and JSON `object`) and emits an `FDv2ChangeSet` without deserializing flags/segments.
> 
> Adds a server-side `TranslateChangeSet` step that converts raw FDv2 changes into typed `ItemDescriptor<Flag|Segment>` changes for application, warning/skip on unknown kinds and aborting the whole changeset on deserialization failure.
> 
> Introduces a generic `data_model::ChangeSet<T>` + shared `ChangeSetType`, updates `MemoryStore::Apply` and `FDv2SourceResult` to consume `ChangeSet<ChangeSetData>`, and wires translation into FDv2 polling (including a new translation-failure interrupted error). Tests are updated and new translator unit tests are added, including the behavior change that unknown kinds now pass through the protocol layer.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 854edbd8aab212c59749f98226076ebc8fe7c414. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


libs/server-sdk/src/data_systems/fdv2/fdv2_changeset_translation.cpp
```diff
@@ -0,0 +1,119 @@
...
+
+    ChangeSetData changes;
+    changes.reserve(change_set.changes.size());
+
+    for (auto const& change : change_set.changes) {
+        if (change.change_type == data_model::FDv2Change::ChangeType::kDelete) {
+            if (auto item = TranslateDelete(change, logger)) {
+                changes.push_back(std::move(*item));
+            }
+        } else {
```
kinyoklion: I think I prefer being explicit in case we add subsequent change types. So we can at least just log the unrecognized change type.

kinyoklion: So, if we add patch, and they accidentally send that to a client that doesn't support it, then we have some breadcrumbs to know what went wrong.

libs/server-sdk/src/data_systems/fdv2/fdv2_changeset_translation.cpp
```diff
@@ -0,0 +1,119 @@
...
+    auto result = boost::json::value_to<
+        tl::expected<std::optional<data_model::Flag>, JsonError>>(
+        change.object);
+    if (!result) {
+        LD_LOG(logger, LogLevel::kError)
+            << "FDv2: could not deserialize flag '" << change.key << "'";
+        return false;
+    }
+    if (!result->has_value()) {
+        LD_LOG(logger, LogLevel::kWarn)
```
kinyoklion, line 53: I think this is a change from a JSON parse error.
```
if (!result->has_value()) {
    return tl::make_unexpected(Error::JsonParseError(
        "flag '" + put.key + "' object was null"));
}
```

Where I still think we  would want the data source level to treat this as malformed JSON for reconnect triggering purposes/status reporting.

libs/server-sdk/src/data_systems/fdv2/fdv2_changeset_translation.cpp
```diff
@@ -0,0 +1,119 @@
...
+        return ItemChange{change.key,
+                          data_model::ItemDescriptor<data_model::Segment>{
+                              data_model::Tombstone{change.version}}};
+    }
+    LD_LOG(logger, LogLevel::kWarn) << "FDv2: unknown kind '" << change.kind
+                                    << "' in delete-object, skipping";
+    return std::nullopt;
+}
+
+static bool TranslatePutFlag(data_model::FDv2Change const& change,
```
kinyoklion: nit: Maybe we could use a generic?
```
template <typename T>
bool TranslatePut(data_model::FDv2Change const& change,
                  char const* kind_name,
                  ChangeSetData* changes,
                  Logger const& logger) { ... }
```

kinyoklion (changes requested):

libs/server-sdk/src/data_systems/fdv2/fdv2_changeset_translation.cpp
beekld: I think this is the same behavior as Java:

https://github.com/launchdarkly/java-core/blob/f4719ecf55d11f2e16a34b904349f98c469213ba/lib/sdk/server/src/main/java/com/launchdarkly/sdk/server/FDv2ChangeSetTranslator.java#L74

I don't really think either one is more "correct" (unless the spec covers this case). But I'm happy to change it if you feel strongly.

libs/server-sdk/src/data_systems/fdv2/fdv2_changeset_translation.cpp
beekld: Heh, I had considered that, and was on the fence about it. Done.

libs/server-sdk/src/data_systems/fdv2/fdv2_changeset_translation.cpp
beekld: Good idea. Done.

libs/server-sdk/src/data_systems/fdv2/fdv2_changeset_translation.cpp
kinyoklion: I am trying to think what really should happen if this case was encountered. I am not sure either is amazing. I am fine with this for the moment.

kinyoklion (approved):

libs/server-sdk/src/data_systems/fdv2/fdv2_changeset_translation.cpp
```diff
@@ -0,0 +1,107 @@
...
+        LD_LOG(logger, LogLevel::kError)
+            << "FDv2: could not deserialize " << kind_name << " '" << change.key
+            << "'";
+        return false;
+    }
+    if (!result->has_value()) {
+        LD_LOG(logger, LogLevel::kWarn)
+            << "FDv2: " << kind_name << " '" << change.key
+            << "' object was null, skipping";
+        return true;
```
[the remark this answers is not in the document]
beekld: Already discussed

PR #526, opened 2026-04-27
refactor: add a hook to SSE client to allow changing the target

beekld:
Added a new hook to the SSE client builder to allow setting a callback function that can modify the target before each connection. The target includes the URL's path and query parameters, but not the host or scheme. This allows changing the selector passed to the backend while still letting the SSE client handle retries itself internally.

I chose the `http::request` object as the thing to be mutable because it's already a public part of the API, and because changing anything else (such as the host or scheme) would be a _much_ more involved change.

Changed the tests to run for both Curl and Foxy, instead of only Curl. This found a couple of minor Foxy bugs, which are fixed.

There's a lot of churn from running clang-format. I apologize it makes the review more difficult, but I think it's the right thing to do. :-/

---

> [!NOTE]
> **Medium Risk**
> Touches connection initiation/reconnect behavior and request construction in both Foxy and CURL backends; incorrect hook interactions or URL/redirect handling could break connectivity or retry semantics.
> 
> **Overview**
> Adds `Builder::on_connect` (a `ConnectionHook`) to let callers mutate the outgoing `http::request` immediately before every connection attempt (initial connect and reconnect), enabling per-connection changes like varying the request target/headers/body.
> 
> Wires the hook through both networking backends: Foxy now invokes the hook before `async_connect`, re-prepares payload, and updates redirect handling to also change `host_`/`port_`; CURL now applies the hook before each transfer, rebuilds the URL from the possibly-mutated target, and explicitly reasserts client-managed `Last-Event-ID` semantics.
> 
> Hardens request construction by ensuring an empty URL path is sent as `/`, and replaces CURL-only tests with a unified `client_test.cpp` suite (including new hook behavior tests).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c7d13d8ab6c00c5180b9b755f2de15fc0ddfcc59. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


libs/server-sent-events/src/client.cpp
```diff
@@ -650,7 +668,9 @@ std::shared_ptr<Client> Builder::build() {
     std::string host = uri_components->host();
 
     request.set(http::field::host, host);
-    request.target(uri_components->encoded_target());
+    // RFC 7230: an empty path in origin-form must be sent as "/".
+    auto target = uri_components->encoded_target();
```
kinyoklion, line 677: Interesting. Were you hitting this?

beekld: Yes, the unit tests set a url without a trailing /, and they pass fine in curl, but they were breaking most of the tests in Foxy

libs/server-sent-events/include/launchdarkly/sse/client.hpp
```diff
@@ -34,6 +34,8 @@ class Builder {
     using EventReceiver = std::function<void(Event)>;
     using LogCallback = std::function<void(std::string)>;
     using ErrorCallback = std::function<void(Error)>;
+    using ConnectionHook =
```
kinyoklion, line 37: Of course in newer C++ we could make this noexcept without a bunch of hoopla, but not in C++17.

So I am going to suggest we potentially have a try-catch around the connection_hook invocations?

Or we just use strong wording on the signature.

libs/server-sent-events/src/client.cpp
kinyoklion, line 217: We may want to say that last-event-id may be overwritten if provided in the request modifier?

kinyoklion: I think this needs to be changed to be done per-request? Or a body mutation would result in an incorrect content-length for the foxy based client. 

```
  TEST(ClientTest, OnConnectHookCanMutateBody) {                                                                                                                                          
      MockSSEServer server;                                                                                                                                                               
      std::string received_body;    
      std::string received_content_length;                                                                                                                                                
      std::mutex received_mutex;                                                                                                                                                        
                                                                                                                                                                                          
      auto port = server.start([&](auto const& req, auto send_response,                                                                                                                   
                                   auto send_sse_event, auto close) {
          {                                                                                                                                                                               
              std::lock_guard lock(received_mutex);                                                                                                                                     
              received_body = req.body();                                                                                                                                                 
              if (auto it = req.find(http::field::content_length);                                                                                                                        
                  it != req.end()) {     
                  received_content_length = std::string(it->value());                                                                                                                     
              }                                                                                                                                                                           
          }                                                                                                                                                                               
          http::response<http::string_body> res{http::status::ok, 11};                                                                                                                    
          res.set(http::field::content_type, "text/event-stream");                                                                                                                        
          res.chunked(true);                                                                                                                                                              
          send_response(res);                                                                                                                                                             
          send_sse_event(SSEFormatter::event("ok"));                                                                                                                                      
          std::this_thread::sleep_for(10ms);                                                                                                                                              
          close();                                                                                                                                                                        
      });                                                                                                                                                                                 
                                                                                                                                                                                        
      IoContextRunner runner;                                                                                                                                                             
      EventCollector collector;
                                                                                                                                                                                          
      std::string const build_time_body = "short";                                                                                                                                      
      std::string const hook_body =                                                                                                                                                       
          "this-body-is-much-longer-than-the-build-time-body";                                                                                                                          
      ASSERT_GT(hook_body.size(), build_time_body.size());                                                                                                                                
                                                                                                                                                                                          
      auto client =                                                                                                                                                                       
          Builder(runner.context().get_executor(),                                                                                                                                        
                  "http://localhost:" + std::to_string(port))                                                                                                                             
              .receiver([&](Event e) { collector.add_event(std::move(e)); })                                                                                                              
              .method(http::verb::post)                                                                                                                                                   
              .body(build_time_body)                                                                                                                                                      
              .on_connect([&](http::request<http::string_body>* req) {                                                                                                                    
                  req->body() = hook_body;                                                                                                                                              
              })                                                                                                                                                                          
              .build();                                                                                                                                                                 
                                                                                                                                                                                          
      client->async_connect();                                                                                                                                                          
      ASSERT_TRUE(collector.wait_for_events(1));
                                                                                                                                                                                          
      {
          std::lock_guard lock(received_mutex);                                                                                                                                           
          EXPECT_EQ(hook_body, received_body)                                                                                                                                           
              << "Server received a truncated body. Content-Length header was '"                                                                                                          
              << received_content_length << "' (hook set body of size "
              << hook_body.size() << ").";                                                                                                                                                
          EXPECT_EQ(std::to_string(hook_body.size()), received_content_length);                                                                                                         
      }                                                                                                                                                                                   
                                                                                                                                                                                          
      SimpleLatch shutdown_latch(1);
      client->async_shutdown([&] { shutdown_latch.count_down(); });                                                                                                                       
      EXPECT_TRUE(shutdown_latch.wait_for(5000ms));                                                                                                                                       
  }   
```

kinyoklion (changes requested):
Requesting changes for body length recalculation. Details in comment.

libs/server-sent-events/include/launchdarkly/sse/client.hpp
beekld: I don't follow. We don't have a try-catch around LogCallback or ErrorCallback. Are you worried about us accidentally throwing from inside the hook?

What would you wanna do if we did? Mark the connection as immediately failed and generate an error?

libs/server-sent-events/src/client.cpp
beekld: Good catch. TBH, I wasn't really thinking about how the body could be changed in the callback, because we're not intending to do that. But it's definitely better to handle that case. Added the test and fixed.

libs/server-sent-events/src/client.cpp
```diff
@@ -185,6 +187,9 @@ class FoxyClient : public Client,
     }
 
     void do_run() {
+        if (connection_hook_) {
+            connection_hook_(&req_);
+        }
```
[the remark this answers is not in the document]
beekld: Fixed.

libs/server-sent-events/src/client.cpp
beekld: Added a comment on the hook setter.

libs/server-sent-events/src/curl_client.cpp
```diff
@@ -77,74 +73,73 @@ void CurlClient::async_connect() {
...
     if (request_context_->is_shutting_down()) {
         return;
     }
 
+    if (connection_hook_) {
+        connection_hook_(&request_context_->req);
+        request_context_->url = build_url(request_context_->req);
+    }
+
+    request_context_->req.prepare_payload();
```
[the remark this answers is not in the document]
beekld: Fixed and test added.

libs/server-sent-events/src/client.cpp
```diff
@@ -185,6 +187,12 @@ class FoxyClient : public Client,
...
+
+        req_.prepare_payload();
+
```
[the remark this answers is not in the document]
beekld: Fixed.

libs/server-sent-events/include/launchdarkly/sse/client.hpp
kinyoklion: I am not super worried bout it. It would just be nice if we could mark all of them noexcept. 

In other SDKs this is a customer facing feature as well, but ours it won't really be. So we probably are fine as is.

kinyoklion (approved):

PR #534, opened 2026-05-14
feat: implement DynamoDBDataSource + tests

beekld:
## Summary

Implements `DynamoDBDataSource`, similar to `RedisDataSource`, using the Proxy Relay schema as per the schema.

```cpp
DynamoDBClientOptions options;
options.region = "us-east-1";
options.endpoint = "http://localhost:8000";
auto src = DynamoDBDataSource::Create("my-table", "prefix", options);
```

## Differences from the Redis source

- `All()` paginates DynamoDB Query responses via `LastEvaluatedKey`; Redis `hgetall` is single-shot.
- AWS SDK `InitAPI`/`ShutdownAPI` are owned by an singleton (`AwsSdkGuard`); Redis++ needs no equivalent.
- Config takes a `DynamoDBClientOptions` struct (region/endpoint/credentials) instead of a single URI string.
- All reads use `ConsistentRead = true` to avoid DynamoDB's default eventually-consistent staleness (same as Java)
- When no prefix is configured, this source uses bare partition-key values (`features`, `$inited`), matching what Relay writes per the spec. The Redis source unconditionally does `prefix + ":" + name`, so an empty Redis prefix would produce leading-colon keys (`:features`) that no Relay deployment writes. That divergence is never exercised — Redis defaults to a non-empty prefix and the Redis tests always pass one, but wanted to note it.

## Test plan

- [x] 24/24 tests green against `amazon/dynamodb-local`
- [x] Unit-only error tests pass without a backend.
- [x] CI Ubuntu job runs against `amazon/dynamodb-local`; macOS / Windows build-only (matches Redis workflow).
- [x] No AWS header leakage in public surface.

---

> [!NOTE]
> **Medium Risk**
> Adds a new DynamoDB-backed data source implementation plus integration tests and CI wiring, and adjusts CMake/AWS SDK build behavior (static vs shared), which could impact build/link outcomes across platforms.
> 
> **Overview**
> Implements a real `DynamoDBDataSource` (replacing the prior scaffold) that reads flags/segments from a Relay-compatible DynamoDB table, including `Get`, `All` (with pagination), and `Initialized` checks, with consistent-read requests and structured error handling for malformed rows.
> 
> Introduces `DynamoDBClientOptions` plus internal helpers (`AwsSdkGuard`, client factory, namespace/prefix utilities) to manage AWS SDK lifecycle and client configuration without leaking AWS headers in the public API.
> 
> Adds a new `tests` target with DynamoDB Local-backed coverage and updates CI/workflows and release scripts to enable DynamoDB builds/tests when requested; also forces the AWS SDK FetchContent build to static archives with cache save/restore to avoid macOS shared-link issues.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c4f97d6a6311decf37fe9ba4d61f687f63825edb. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/libs/server-sdk-dynamodb-source/src/aws_sdk_guard.hpp
+++ b/libs/server-sdk-dynamodb-source/src/aws_sdk_guard.hpp
@@ -1,0 +1,38 @@
+#pragma once
+
+#include <aws/core/Aws.h>
+
+namespace launchdarkly::server_side::integrations::detail {
+
+// AwsSdkGuard owns the process-wide Aws::InitAPI / Aws::ShutdownAPI lifecycle
+// for this library. Multiple DynamoDB-backed integrations within the same
+// process share the single static instance; the API is initialized lazily on
+// first use and torn down during normal program termination via C++ static
+// destruction.
+//
+// Static-destruction ordering caveat: if a caller stashes a raw AWS SDK
+// pointer in their own static and that static is destroyed AFTER this guard,
+// AWS SDK calls during that destructor will be undefined. The standard usage
+// pattern (holding the data source / store via a unique_ptr or shared_ptr in
+// regular program scope, not in another static) is unaffected because those
+// smart pointers destruct before the guard.
+class AwsSdkGuard {
+   public:
+    // Idempotent. First call constructs the singleton, which runs
+    // Aws::InitAPI in its constructor. Subsequent calls are no-ops. Safe to
+    // call from any thread.
+    static void Ensure();
+
+    AwsSdkGuard(AwsSdkGuard const&) = delete;
+    AwsSdkGuard(AwsSdkGuard&&) = delete;
+    AwsSdkGuard& operator=(AwsSdkGuard const&) = delete;
+    AwsSdkGuard& operator=(AwsSdkGuard&&) = delete;
+
+   private:
+    AwsSdkGuard();
+    ~AwsSdkGuard();
+
+    Aws::SDKOptions options_;
+};
+
+}  // namespace launchdarkly::server_side::integrations::detail
--- a/libs/server-sdk-dynamodb-source/src/dynamodb_attributes.hpp
+++ b/libs/server-sdk-dynamodb-source/src/dynamodb_attributes.hpp
@@ -1,0 +1,22 @@
+#pragma once
+
+namespace launchdarkly::server_side::integrations::detail {
+
+// Schema constants for the LaunchDarkly DynamoDB table layout, shared with
+// other DynamoDB-backed integrations in this library.
+//
+// The schema is defined in sdk-specs/specs/PS-persistent-store/README.md §
+// DynamoDB schema and is the same layout that the LaunchDarkly Relay Proxy
+// writes; mismatching any of these strings means the SDK and Relay will not
+// agree on what's in the table.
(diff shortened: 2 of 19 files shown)


libs/server-sdk-dynamodb-source/src/dynamodb_attributes.hpp
```diff
@@ -0,0 +1,22 @@
...
+// Schema constants for the LaunchDarkly DynamoDB table layout, shared with
+// other DynamoDB-backed integrations in this library.
+//
+// The schema is defined in sdk-specs/specs/PS-persistent-store/README.md §
+// DynamoDB schema and is the same layout that the LaunchDarkly Relay Proxy
+// writes; mismatching any of these strings means the SDK and Relay will not
+// agree on what's in the table.
+inline constexpr char kPartitionKey[] = "namespace";
+inline constexpr char kSortKey[] = "key";
+inline constexpr char kVersionAttribute[] = "version";
```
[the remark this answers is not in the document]
beekld: fixed

libs/server-sdk-dynamodb-source/src/dynamodb_source.cpp
```diff
@@ -1,15 +1,154 @@
...
+        return std::nullopt;
+    }
+
+    auto const it = item.find(kItemAttribute);
+    if (it == item.end()) {
+        return tl::make_unexpected(
+            Error{"DynamoDB row missing expected 'item' attribute"});
+    }
+
+    return SerializedItemDescriptor::Present(0, it->second.GetS());
```
kinyoklion: I wonder if we should type check or empty check calls to GetS? I have claude working on a test so I can understand scope a little easier.

libs/server-sdk-dynamodb-source/src/aws_sdk_guard.hpp
```diff
@@ -0,0 +1,38 @@
...
+
+namespace launchdarkly::server_side::integrations::detail {
+
+// AwsSdkGuard owns the process-wide Aws::InitAPI / Aws::ShutdownAPI lifecycle
+// for this library. Multiple DynamoDB-backed integrations within the same
+// process share the single static instance; the API is initialized lazily on
+// first use and torn down during normal program termination via C++ static
+// destruction.
+//
+// Static-destruction ordering caveat: if a caller stashes a raw AWS SDK
```
kinyoklion, line 13: Do we need to surface this someplace more customer visible? Readme? Or just when we make docs?

kinyoklion: I guess I need to read further to understand if there is external impact. On some SDKs we allow passing clients, but they would then theoretically be non-owned. So more of an internal problem?

kinyoklion: Oh, the AWS Api is a bit icky here.

kinyoklion: I wonder if potentially we should have a way to disable the AWS SDK shutdown and let the customer be responsible for shutting it down?

kinyoklion (approved):

PR #536, opened 2026-05-19
feat: add IBigSegmentStore interface + Redis and DynamoDB stores

beekld:
Ticket: SDK-2363 · Follows #534

## Summary

Adds the public `IBigSegmentStore` interface, the `Membership` / `StoreMetadata` value types, and concrete DynamoDB + Redis implementations. Schema strings match what the Relay Proxy writes.

```cpp
auto redis_store = RedisBigSegmentStore::Create("redis://localhost:6379", "prefix");
auto dynamo_store = DynamoDBBigSegmentStore::Create("my-table", "prefix", options);
```

## Design notes

- **`synchronizedOn` parsing.** Stored as DynamoDB N / Redis string. Both stores reject malformed values (non-numeric strings, wrong DynamoDB attribute type) rather than silently returning 0, matching the existing `dynamodb_source.cpp` row validation.
- **No hashing here.** The interface contract says callers pass the already-hashed base64 SHA-256 context key; the SDK will hash in the wrapper, not the store implementations.

## Not in scope

- `BigSegmentsBuilder` config plumbing.
- `BigSegmentStoreWrapper` (LRU cache + staleness polling) and the hashing path.
- Evaluator wiring / replacing the `rules.cpp` big-segments TODO.

## Test plan

- [x] 7 `Membership` unit tests pass.
- [x] 9 `RedisBigSegmentStore` integration tests pass against Redis 7 (docker).
- [x] 8 `DynamoDBBigSegmentStore` integration tests pass against DynamoDB Local.
- [x] Full server-sdk test suite (468 tests) still green.
- [x] Schema constants match what Relay writes (verified against the LocalStack/Redis fixture data).

---

> [!NOTE]
> **Medium Risk**
> New external-store read path for segment targeting with strict schema validation; not yet wired into evaluation, so production flag behavior is unchanged until follow-up PRs land.
> 
> **Overview**
> Introduces a **Big Segments persistent store** surface for the server SDK: public `IBigSegmentStore`, inline `Membership` / `StoreMetadata`, and read-only **Redis** and **DynamoDB** backends that follow Relay’s key/schema layout (prefix scoping, include/exclude refs, sync metadata).
> 
> Stores perform point lookups by opaque context hash, build membership via `Membership::FromSegmentRefs` (inclusion wins on overlap), and return errors on malformed DynamoDB attribute types or non-numeric sync timestamps instead of silent empty/zero results. **Config, caching wrapper, hashing, and flag evaluation wiring are not included** in this change.
> 
> Integration tests cover membership/metadata behavior, prefix isolation, and error paths against Redis and DynamoDB Local.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d2622d9be099d458afba60bf1a58e7f948af8859. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


libs/server-sdk-dynamodb-source/src/dynamodb_big_segment_store.cpp
```diff
@@ -0,0 +1,148 @@
...
+        it != item.end()) {
+        for (auto const& ref : it->second.GetSS()) {
+            included.emplace_back(ref);
+        }
+    }
+    if (auto const it = item.find(kBigSegmentsExcludedAttribute);
+        it != item.end()) {
+        for (auto const& ref : it->second.GetSS()) {
+            excluded.emplace_back(ref);
+        }
```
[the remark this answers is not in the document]
beekld: Fixed

libs/server-sdk-dynamodb-source/src/dynamodb_big_segment_store.cpp
```diff
@@ -0,0 +1,164 @@
...
+                std::string("DynamoDB Big Segments '") +
+                kBigSegmentsExcludedAttribute +
+                "' is not of type STRING_SET");
+        }
+        for (auto const& ref : it->second.GetSS()) {
+            excluded.emplace_back(ref);
+        }
+    }
+
+    return Membership::FromSegmentRefs(included, excluded);
```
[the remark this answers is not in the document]
beekld: Fixed.

libs/server-sdk-redis-source/src/redis_big_segment_store.cpp
```diff
@@ -0,0 +1,90 @@
...
+
+    try {
+        redis_->smembers(include_key, std::back_inserter(included));
+        redis_->smembers(exclude_key, std::back_inserter(excluded));
+    } catch (sw::redis::Error const& e) {
+        return tl::make_unexpected(e.what());
+    }
+
+    return Membership::FromSegmentRefs(included, excluded);
+}
```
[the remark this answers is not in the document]
beekld: Fixed.

libs/server-sdk-redis-source/src/redis_big_segment_store.cpp
```diff
@@ -0,0 +1,90 @@
...
+
+}  // namespace
+
+tl::expected<std::unique_ptr<RedisBigSegmentStore>, std::string>
+RedisBigSegmentStore::Create(std::string uri, std::string prefix) {
+    try {
+        return std::unique_ptr<RedisBigSegmentStore>(new RedisBigSegmentStore(
+            std::make_unique<sw::redis::Redis>(std::move(uri)),
+            std::move(prefix)));
+    } catch (sw::redis::Error const& e) {
```
kinyoklion, line 42: Should we catch std::exception as well?

libs/server-sdk-dynamodb-source/tests/dynamodb_big_segment_store_test.cpp
```diff
@@ -0,0 +1,164 @@
...
+    options.aws_access_key_id = "dummy";
+    options.aws_secret_access_key = "dummy";
+    return options;
+}
+
+class DynamoDBBigSegmentTests : public ::testing::Test {
+   public:
+    DynamoDBBigSegmentTests()
+        : table_name_("ld-dynamodb-big-segments-test"),
+          prefix_("testprefix"),
```
kinyoklion, line 44: Should we have a test for an empty prefix?

libs/server-sdk-redis-source/src/redis_big_segment_store.cpp
```diff
@@ -0,0 +1,90 @@
...
+        return tl::make_unexpected(e.what());
```
kinyoklion, line 43: Do we need to sanitize this for the redis user/password?

libs/server-sdk-redis-source/src/redis_big_segment_store.cpp
```diff
@@ -0,0 +1,90 @@
...
+    : redis_(std::move(redis)),
+      prefix_(std::move(prefix)),
+      sync_time_key_(prefix_ + ":" + kSyncTimeKey) {}
+
+RedisBigSegmentStore::~RedisBigSegmentStore() = default;
+
+IBigSegmentStore::GetMembershipResult RedisBigSegmentStore::GetMembership(
+    std::string const& context_hash) const {
+    std::string const include_key =
+        prefix_ + ":" + kIncludeKeyNamespace + ":" + context_hash;
```
kinyoklion: nit: Would it be work it to cache `prefix_ + ":" + kIncludeKeyNamespace` and `prefix_ + ":" + kExcludeKeyNamespace` then just add on the context_hash?

libs/server-sdk-redis-source/src/redis_big_segment_store.cpp
```diff
@@ -0,0 +1,90 @@
...
+        return tl::make_unexpected(e.what());
+    }
+}
+
+RedisBigSegmentStore::RedisBigSegmentStore(
+    std::unique_ptr<sw::redis::Redis> redis,
+    std::string prefix)
+    : redis_(std::move(redis)),
+      prefix_(std::move(prefix)),
+      sync_time_key_(prefix_ + ":" + kSyncTimeKey) {}
```
kinyoklion: Should we check for an empty prefix? 

.Net and JS do, but Java doesn't.

```csharp
public RedisStoreBuilder<T> Prefix(string prefix)
{
    _prefix = string.IsNullOrEmpty(prefix) ? Redis.DefaultPrefix : prefix;
    return this;
}
```

libs/server-sdk-redis-source/src/redis_big_segment_store.cpp
kinyoklion: Or maybe make this and the dynamo version into something vague? Main concern is just a URI coming out which includes user/password or something else the customer puts into the URI and considers secret.

libs/server-sdk-dynamodb-source/src/dynamodb_big_segment_store.cpp
```diff
@@ -0,0 +1,161 @@
...
+    }
+
+    auto const& item = outcome.GetResult().GetItem();
+    if (item.empty()) {
+        return std::nullopt;
+    }
+
+    auto const it = item.find(kBigSegmentsSyncTimeAttribute);
+    if (it == item.end()) {
+        return tl::make_unexpected(
```
kinyoklion: Other implementations seem more lenient here. Though I wonder if there was an evolution where maybe this field didn't exist originally.

kinyoklion (changes requested):

libs/server-sdk-redis-source/src/redis_big_segment_store.cpp
beekld: Done

libs/server-sdk-dynamodb-source/tests/dynamodb_big_segment_store_test.cpp
beekld: Done

libs/server-sdk-redis-source/src/redis_big_segment_store.cpp
beekld: Done

libs/server-sdk-redis-source/src/redis_big_segment_store.cpp
beekld: Done

Yes, I think Java is probably wrong here

libs/server-sdk-dynamodb-source/src/dynamodb_big_segment_store.cpp
beekld: The spec seems to say we should be more lenient too. Done.

libs/server-sdk-redis-source/src/redis_big_segment_store.cpp
beekld: I wanna push back on this one. I think making it vague would lose the diagnostic value of the original error messages. I would hope that redis already wouldn't be putting sensitive info into error messages. But also, we aren't logging this here, either. I think the "correct" solution would be to have a sanitizer wherever we are logging, but that's a much bigger task.

kinyoklion (approved):

PR #537, opened 2026-05-21
refactor: add SSE Builder::on_response hook

beekld:
## Summary

Adds an `on_response` hook to `sse::Builder`, invoked once per (re)connect after the HTTP response headers have been received, before any branching on status. Fires for every response (any status, including redirects and errors); does not fire if the connection failed before any response (TCP error, read timeout).

Prep for upcoming FDv2 work that needs to inspect the `X-LD-FD-Fallback` response header.

## Design notes

cURL backend accumulates header lines into a per-transfer `http::response_header<>` on `RequestContext`, because `HeaderCallback` is invoked once per CRLF-terminated line. The hook fires on the empty terminator and is posted to the client strand.

## Test plan

- [x] `gtest_launchdarkly-sse-client` — green under both Foxy and cURL backends
- [x] New `ClientTest.OnResponseHookFiresWithCustomHeader`

---

> [!NOTE]
> **Medium Risk**
> Touches reconnect/header parsing on both Foxy and cURL SSE paths; cURL header assembly is new logic though covered by existing and new client tests.
> 
> **Overview**
> Adds **`sse::Builder::on_response`**, a callback that runs **once per connect/reconnect** right after response headers are parsed and **before** status-based handling (success, redirect, errors). It receives a `http::response_header<>` so callers can read headers such as **`X-LD-FD-Fallback`** without waiting for the SSE body.
> 
> **Foxy** backend invokes the hook synchronously on the client executor when header parsing completes. **cURL** backend rebuilds the prior Content-Type-only header sniffing into full **line-by-line header assembly** (status line, OWS trimming, duplicate headers via `insert()`, redirect chains), emits the hook on the blank line after headers, and **posts** it to the same executor as other callbacks. Header accumulator state is reset on each new transfer.
> 
> Includes **`ClientTest.OnResponseHookFiresWithCustomHeader`** for status and custom header visibility.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 733ec5121aec64f85c4072d21077411bdc1fa23a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/libs/server-sent-events/src/curl_client.cpp
+++ b/libs/server-sent-events/src/curl_client.cpp
@@ -3,10 +3,13 @@
 #include "curl_client.hpp"
 
 #include <boost/asio/post.hpp>
+#include <boost/beast/core/string.hpp>
 #include <boost/beast/http/status.hpp>
 #include <boost/url/url.hpp>
 
+#include <charconv>
 #include <sstream>
+#include <system_error>
 
 namespace launchdarkly::sse {
 namespace beast = boost::beast;
@@ -41,6 +44,7 @@ CurlClient::CurlClient(
     Builder::LogCallback logger,
     Builder::ErrorCallback errors,
     Builder::ConnectionHook connection_hook,
+    Builder::ResponseHook response_hook,
     bool skip_verify_peer,
     std::optional<std::string> custom_ca_file,
     bool use_https,
@@ -51,6 +55,7 @@ CurlClient::CurlClient(
       logger_(std::move(logger)),
       errors_(std::move(errors)),
       connection_hook_(std::move(connection_hook)),
+      response_hook_(std::move(response_hook)),
       use_https_(use_https),
       backoff_timer_(executor),
       multi_manager_(CurlMultiManager::create(executor)),
@@ -147,6 +152,19 @@ void CurlClient::do_run() {
             if (auto ctx = weak_ctx.lock()) {
                 if (auto const self = weak_self.lock()) {
                     self->log_message(message);
+                }
+            }
+        },
+        [weak_self, weak_ctx](http::response_header<> headers) {
+            if (auto ctx = weak_ctx.lock()) {
+                if (auto const self = weak_self.lock()) {
+                    if (self->response_hook_) {
+                        boost::asio::post(
+                            self->backoff_timer_.get_executor(),
+                            [self, headers = std::move(headers)]() {
+                                self->response_hook_(headers);
+                            });
+                    }
                 }
             }
         }));
@@ -397,6 +415,11 @@ size_t CurlClient::WriteCallback(char const* data,
 // Callback for reading request headers
 //
 // https://curl.se/libcurl/c/CURLOPT_HEADERFUNCTION.html
+//
+// libcurl invokes this once per CRLF-terminated response line: the HTTP status
+// line, then each header, then an empty terminator. With
+// CURLOPT_FOLLOWLOCATION enabled the cycle repeats for each response in the
+// redirect chain.
 size_t CurlClient::HeaderCallback(char const* buffer,
                                   size_t size,
                                   size_t nitems,
@@ -404,12 +427,53 @@ size_t CurlClient::HeaderCallback(char const* buffer,
     size_t const total_size = size * nitems;
     auto* context = static_cast<RequestContext*>(userdata);
 
-    // Check for Content-Type header
-    if (std::string const header(buffer, total_size);
-        header.find("Content-Type:") == 0 ||
-        header.find("content-type:") == 0) {
-        if (header.find("text/event-stream") == std::string::npos) {
-            context->log_message("warning: unexpected Content-Type: " + header);
+    std::string_view line(buffer, total_size);
+    if (line.size() >= 2 && line[line.size() - 2] == '\r' &&
+        line.back() == '\n') {
+        line.remove_suffix(2);
+    }
+
+    if (line.empty()) {
+        // End-of-headers terminator: emit and reset.
+        context->response(std::move(context->current_response));
+        context->current_response = http::response_header<>{};
+        return total_size;
+    }
+
+    if (line.substr(0, 5) == "HTTP/") {
+        // Status line: "HTTP/X.Y CODE REASON". Start a fresh response.
+        context->current_response = http::response_header<>{};
+        auto const code_start = line.find(' ');
+        if (code_start != std::string_view::npos) {
+            unsigned code = 0;
+            auto const result = std::from_chars(
+                line.data() + code_start + 1, line.data() + line.size(), code);
+            if (result.ec == std::errc{} && code != 0) {
+                context->current_response.result(code);
+            }
+        }
+        return total_size;
+    }
+
+    auto const colon = line.find(':');
+    if (colon != std::string_view::npos) {
+        std::string_view name = line.substr(0, colon);
+        // HTTP optional whitespace (OWS) per RFC 7230 §3.2.3 is SP or HTAB.
+        std::string_view value = line.substr(colon + 1);
+        while (!value.empty() &&
+               (value.front() == ' ' || value.front() == '\t')) {
+            value.remove_prefix(1);
+        }
+        while (!value.empty() &&
+               (value.back() == ' ' || value.back() == '\t')) {
+            value.remove_suffix(1);
+        }
+        context->current_response.set(std::string(name), std::string(value));
+
+        if (beast::iequals(name, "Content-Type") &&
+            value.find("text/event-stream") == std::string_view::npos) {
+            context->log_message("warning: unexpected Content-Type: " +
+                                 std::string(line));
         }
     }
 
--- a/libs/server-sent-events/src/curl_client.cpp
+++ b/libs/server-sent-events/src/curl_client.cpp
@@ -516,6 +516,11 @@ void CurlClient::PerformRequestWithMulti(
     // Initialize parser for new connection (last_event_id is tracked
     // separately)
     context->init_parser();
+    // Reset header-accumulator state in case the previous transfer dropped
+    // mid-headers, which would otherwise leave reading_headers=true and
+    // cause the new response's HTTP/ status line to be skipped.
+    context->current_response = http::response_header<>{};
+    context->reading_headers = false;
 
     std::shared_ptr<CURL> curl(curl_easy_init(), curl_easy_cleanup);
     if (!curl) {
--- a/libs/server-sent-events/src/curl_client.cpp
+++ b/libs/server-sent-events/src/curl_client.cpp
@@ -437,12 +437,32 @@ size_t CurlClient::HeaderCallback(char const* buffer,
         // End-of-headers terminator: emit and reset.
         context->response(std::move(context->current_response));
         context->current_response = http::response_header<>{};
+    // Strip the line terminator. Allow bare LF or bare CR per RFC 9112 §2.2;
+    // libcurl preserves the original wire bytes for HTTP/1.x (only HTTP/2
+    // synthesizes CRLF), so a non-compliant origin can deliver bare LF here.
+    while (!line.empty() &&
+           (line.back() == '\r' || line.back() == '\n')) {
+        line.remove_suffix(1);
+    }
+
+    if (line.empty()) {
+        // Terminator. If we're between responses (e.g., the line ends a
+        // chunked-transfer trailer block), there's nothing to emit.
+        if (context->reading_headers) {
+            context->response(std::move(context->current_response));
+            context->current_response = http::response_header<>{};
+            context->reading_headers = false;
+        }
         return total_size;
     }
 
     if (line.substr(0, 5) == "HTTP/") {
-        // Status line: "HTTP/X.Y CODE REASON". Start a fresh response.
-        // Status line: "HTTP/X.Y CODE REASON". Start a fresh response.
+        // Status line: "HTTP/X.Y CODE REASON". Only legitimate before any
+        // header has been seen for this response — an interior HTTP/ line
+        // would otherwise wipe accumulated state.
+        if (context->reading_headers) {
+            return total_size;
+        }
         // Beast default-constructs result_ to status::ok (200); reset to 0
         // so an unparseable status line surfaces as result_int() == 0.
         context->current_response = http::response_header<>{};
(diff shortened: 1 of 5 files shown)


libs/server-sent-events/src/curl_client.cpp
```diff
@@ -397,19 +415,65 @@ size_t CurlClient::WriteCallback(char const* data,
...
+    if (line.empty()) {
+        // End-of-headers terminator: emit and reset.
+        context->response(std::move(context->current_response));
+        context->current_response = http::response_header<>{};
+        return total_size;
+    }
+
+    if (line.substr(0, 5) == "HTTP/") {
+        // Status line: "HTTP/X.Y CODE REASON". Start a fresh response.
+        context->current_response = http::response_header<>{};
```
kinyoklion, line 459: **H2 — Malformed status line silently leaves `result_int() == 200`** *(minimal fix; skip if applying the comprehensive M1+M3 fix below).*

Beast default-constructs `response_header<>` with `result_ = status::ok` (verified at Beast `message.hpp:439`). Combined with `std::from_chars` rejecting any leading whitespace, an unparseable status line like `HTTP/1.1  200 OK` (double-space — invalid HTTP grammar but emitted by some misbehaving proxies) or `HTTP/1.1\r\n` (no code) silently leaves `result_int() == 200`. The `code != 0` guard two lines down never fires, so the hook receives a manufactured 200 status for what was actually an unparseable response.

The FDv2 fallback heuristic motivating this PR will misclassify a malformed response as a healthy 200. Explicitly resetting to 0 (=`status::unknown`) makes parse failures visible to callers via the existing `result_int()` channel.

```suggestion
        // Status line: "HTTP/X.Y CODE REASON". Start a fresh response.
        // Beast default-constructs result_ to status::ok (200); reset to 0
        // so an unparseable status line surfaces as result_int() == 0.
        context->current_response = http::response_header<>{};
        context->current_response.result(0);
```

libs/server-sent-events/src/curl_client.cpp
```diff
@@ -397,19 +415,65 @@ size_t CurlClient::WriteCallback(char const* data,
...
+        auto const code_start = line.find(' ');
+        if (code_start != std::string_view::npos) {
+            unsigned code = 0;
+            auto const result = std::from_chars(
+                line.data() + code_start + 1, line.data() + line.size(), code);
+            if (result.ec == std::errc{} && code != 0) {
```
kinyoklion: **H1 — `response_header<>::result(unsigned)` throws on `code > 999`; thrown across the libcurl C frame → UB** *(minimal fix; skip if applying the comprehensive M1+M3 fix below).*

Beast `message.hpp:372`: `@throws std::invalid_argument if v > 999.`

`HeaderCallback` runs inside `curl_multi_perform`, a C frame. Throwing a C++ exception across a C frame is undefined behavior — in practice the process terminates. A server (or transparent proxy with naive re-stamping) emitting `HTTP/1.1 1234 ...` is enough to crash the SDK: `from_chars` parses 4 ASCII digits cleanly, the `code != 0` guard passes, and `result(1234)` throws.

Tightening the guard to the valid HTTP status range (RFC 7231 §6: status codes are three-digit) avoids the throw.

```suggestion
            if (result.ec == std::errc{} && code >= 100 && code <= 999) {
```

libs/server-sent-events/src/curl_client.cpp
```diff
@@ -397,19 +415,65 @@ size_t CurlClient::WriteCallback(char const* data,
...
+        std::string_view value = line.substr(colon + 1);
+        while (!value.empty() &&
+               (value.front() == ' ' || value.front() == '\t')) {
+            value.remove_prefix(1);
+        }
+        while (!value.empty() &&
+               (value.back() == ' ' || value.back() == '\t')) {
+            value.remove_suffix(1);
+        }
+        context->current_response.set(std::string(name), std::string(value));
```
kinyoklion: **H3 — `set()` collapses duplicate-name headers; introduces a silent backend asymmetry** *(minimal fix; skip if applying the comprehensive M1+M3 fix below).*

Beast's `basic_fields::set()` semantics: "Set a field value, removing any other instances of that field." For headers that legitimately repeat — `Set-Cookie` (RFC 6265 §3 forbids merging), `Via`, `Warning`, `Link`, `Proxy-Authenticate`, `WWW-Authenticate`, list-valued headers sent as separate lines — only the LAST value survives in `current_response`.

The Foxy backend uses Beast's parser (`insert()` semantics) and preserves all duplicates. Same server response → different `response_header<>` delivered to the hook depending on which backend is compiled. Today's hook caller is FDv2 reading a single-valued `X-LD-FD-Fallback`, but the public `ResponseHook` signature gives callers the full `response_header<>` and the contract makes no statement about lossy collapsing.

```suggestion
        context->current_response.insert(std::string(name), std::string(value));
```

libs/server-sent-events/src/curl_client.hpp
```diff
@@ -94,6 +97,10 @@ class CurlClient final : public Client,
         std::chrono::steady_clock::time_point last_progress_time;
         curl_off_t last_download_amount;
 
+        // Accumulator for the current response's headers; reset on each new
+        // status line, emitted on the empty terminator line.
+        http::response_header<> current_response;
```
kinyoklion, line 102: **M1 + M3 — parser state flag (companion to the cpp suggestion below).** *(Comprehensive fix path — apply this with the M1+M3 cpp suggestion. Skip if applying the minimal H1/H2/H3 one-liners above.)*

M1 (chunked trailers fire a phantom second hook) and M3 (interior `HTTP/` line resets accumulated state) share a root cause: `HeaderCallback` has no concept of "mid-response vs between responses." A single state bit on `RequestContext` is enough to gate both.

```suggestion
        http::response_header<> current_response;

        // True while accumulating headers between an `HTTP/` status line and
        // the empty terminator. Gates `HeaderCallback` against chunked
        // trailers (which arrive without a fresh status line) and against
        // interior `HTTP/` lines that would otherwise wipe accumulated state.
        bool reading_headers = false;
```

libs/server-sent-events/src/curl_client.cpp
```diff
@@ -397,19 +415,65 @@ size_t CurlClient::WriteCallback(char const* data,
...
+
+        if (beast::iequals(name, "Content-Type") &&
+            value.find("text/event-stream") == std::string_view::npos) {
+            context->log_message("warning: unexpected Content-Type: " +
+                                 std::string(line));
         }
     }
```
kinyoklion, line 511: **M1 + M3 — comprehensive HeaderCallback rewrite (bundles H1, H2, H3, M1, M3, and L1/§2.2 tolerance).** *(Comprehensive fix path — apply this with the hpp suggestion above. Skip if applying the minimal H1/H2/H3 one-liners individually; the line ranges conflict.)*

This introduces a single `reading_headers` state bit and gates the three branches on it:

- **M1 (trailers)** — Per libcurl's [`CURLOPT_HEADERFUNCTION` docs](https://curl.se/libcurl/c/CURLOPT_HEADERFUNCTION.html), the callback is invoked for chunked trailers and their terminator line. Without a state flag, the trailer block accumulates into a freshly-reset `current_response` (status 200 by Beast default) and fires the hook a second time. Gating both the terminator and the trailer headers on `reading_headers` suppresses the phantom fire.

- **M3 (interior `HTTP/` reset)** — A non-leading `HTTP/...` line currently wipes accumulated state. The Foxy backend doesn't have this issue (Beast rejects interior status lines as a protocol error); aligning the curl backend closes the gap.

- **L1 / RFC 9112 §2.2 tolerance** — The existing strip at lines 431-434 only removes paired CRLF, so a bare-LF or bare-CR terminator leaves a stray byte in the value. Verified against libcurl source: for HTTP/1.x, libcurl forwards the original wire bytes to HEADERFUNCTION (`lib/http.c:4456,4490-4492`), so a non-compliant origin can deliver bare LF here. Only HTTP/2 is normalized (`lib/http2.c:1532-1565` synthesizes CRLF). A permissive `while` loop at the top of the suggestion handles bare LF, bare CR, and CRLF uniformly per RFC 9112 §2.2.

- **H1, H2, H3** are folded in inline.

**Manual follow-up after applying this suggestion:** the original 4-line CRLF strip at lines 431-434 (just above this suggestion's range) becomes dead code — the new `while` loop subsumes it. Delete those lines:

```cpp
    if (line.size() >= 2 && line[line.size() - 2] == '\r' &&
        line.back() == '\n') {
        line.remove_suffix(2);
    }
```

```suggestion
    // Strip the line terminator. Allow bare LF or bare CR per RFC 9112 §2.2;
    // libcurl preserves the original wire bytes for HTTP/1.x (only HTTP/2
    // synthesizes CRLF), so a non-compliant origin can deliver bare LF here.
    while (!line.empty() &&
           (line.back() == '\r' || line.back() == '\n')) {
        line.remove_suffix(1);
    }

    if (line.empty()) {
        // Terminator. If we're between responses (e.g., the line ends a
        // chunked-transfer trailer block), there's nothing to emit.
        if (context->reading_headers) {
            context->response(std::move(context->current_response));
            context->current_response = http::response_header<>{};
            context->reading_headers = false;
        }
        return total_size;
    }

    if (line.substr(0, 5) == "HTTP/") {
        // Status line: "HTTP/X.Y CODE REASON". Only legitimate before any
        // header has been seen for this response — an interior HTTP/ line
        // would otherwise wipe accumulated state.
        if (context->reading_headers) {
            return total_size;
        }
        // Beast default-constructs result_ to status::ok (200); reset to 0
        // so an unparseable status line surfaces as result_int() == 0.
        context->current_response = http::response_header<>{};
        context->current_response.result(0);
        auto const code_start = line.find(' ');
        if (code_start != std::string_view::npos) {
            unsigned code = 0;
            auto const result = std::from_chars(
                line.data() + code_start + 1, line.data() + line.size(), code);
            // Three-digit status per RFC 7231 §6; the tight bound avoids
            // result(unsigned) throwing across the libcurl C frame.
            if (result.ec == std::errc{} && code >= 100 && code <= 999) {
                context->current_response.result(code);
            }
        }
        context->reading_headers = true;
        return total_size;
    }

    if (!context->reading_headers) {
        // Header line received outside an active response — chunked trailer
        // or protocol-level junk. Ignore.
        return total_size;
    }

    auto const colon = line.find(':');
    if (colon != std::string_view::npos) {
        std::string_view name = line.substr(0, colon);
        // HTTP optional whitespace (OWS) per RFC 7230 §3.2.3 is SP or HTAB.
        std::string_view value = line.substr(colon + 1);
        while (!value.empty() &&
               (value.front() == ' ' || value.front() == '\t')) {
            value.remove_prefix(1);
        }
        while (!value.empty() &&
               (value.back() == ' ' || value.back() == '\t')) {
            value.remove_suffix(1);
        }
        // insert() preserves duplicate-name headers (Set-Cookie, Via, …);
        // set() would collapse them and diverge from the Foxy backend.
        context->current_response.insert(std::string(name), std::string(value));

        if (beast::iequals(name, "Content-Type") &&
            value.find("text/event-stream") == std::string_view::npos) {
            context->log_message("warning: unexpected Content-Type: " +
                                 std::string(line));
        }
    }
```

libs/server-sent-events/src/curl_client.cpp
```diff
@@ -397,19 +415,65 @@ size_t CurlClient::WriteCallback(char const* data,
...
     auto* context = static_cast<RequestContext*>(userdata);
 
-    // Check for Content-Type header
-    if (std::string const header(buffer, total_size);
-        header.find("Content-Type:") == 0 ||
-        header.find("content-type:") == 0) {
-        if (header.find("text/event-stream") == std::string::npos) {
-            context->log_message("warning: unexpected Content-Type: " + header);
+    std::string_view line(buffer, total_size);
+    if (line.size() >= 2 && line[line.size() - 2] == '\r' &&
```
kinyoklion: I have concerns with this working with RFC 9112 2.2:
```
Although the line terminator for the start-line and fields is the sequence CRLF, a recipient MAY recognize a single LF as a line terminator and ignore any preceding CR.
```
The beast code does apply this lenience, but I am checking now if maybe curl normalizes this already. If not it will be in the blob below.

kinyoklion: Looks like CURL would only normalize it for HTTP/2 where it synthesizes http/1 lines.

beekld: I think this was handled by one of the other suggestions.

libs/server-sent-events/src/curl_client.cpp
beekld: I went with the comprehensive fix.

libs/server-sent-events/src/curl_client.cpp
beekld: I went with the comprehensive fix.

libs/server-sent-events/src/curl_client.cpp
```diff
@@ -397,19 +415,91 @@ size_t CurlClient::WriteCallback(char const* data,
...
+        }
+        return total_size;
+    }
+
+    if (line.substr(0, 5) == "HTTP/") {
+        // Status line: "HTTP/X.Y CODE REASON". Only legitimate before any
+        // header has been seen for this response — an interior HTTP/ line
+        // would otherwise wipe accumulated state.
+        if (context->reading_headers) {
+            return total_size;
```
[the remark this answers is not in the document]
beekld: Fixed.

libs/server-sent-events/src/curl_client.hpp
```diff
@@ -238,6 +261,7 @@ class CurlClient final : public Client,
                Builder::LogCallback logger,
                Builder::ErrorCallback errors,
                Builder::ConnectionHook connection_hook,
+               Builder::ResponseHook response_hook,
```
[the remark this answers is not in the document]
beekld: Fixed.

libs/server-sent-events/src/curl_client.cpp
```diff
@@ -397,19 +415,91 @@ size_t CurlClient::WriteCallback(char const* data,
...
+               (value.front() == ' ' || value.front() == '\t')) {
+            value.remove_prefix(1);
+        }
+        while (!value.empty() &&
+               (value.back() == ' ' || value.back() == '\t')) {
+            value.remove_suffix(1);
+        }
+        // insert() preserves duplicate-name headers (Set-Cookie, Via, …);
+        // set() would collapse them and diverge from the Foxy backend.
+        context->current_response.insert(std::string(name), std::string(value));
```
kinyoklion: This one could still throw for invalid header names.

kinyoklion (approved):

libs/server-sent-events/src/curl_client.cpp
beekld: Fixed.

PR #538, opened 2026-05-23
feat: propagate FDv1 fallback directive from FDv2 polling and streaming sources

beekld:
## Summary

Each FDv2 source now inspects the `X-LD-FD-Fallback` response header (value `"true"`, case-insensitive) and stamps `FDv2SourceResult.fdv1_fallback` on every emitted result. The orchestrator does not yet consume the flag — that comes in a follow-up.

## Design notes

- `fdv1_fallback` is hoisted from per-variant (was on `ChangeSet`, `Interrupted`, `TerminalError`, `Goodbye`) to a top-level field on `FDv2SourceResult`. The flag is a property of the HTTP response, not of the result kind. Touches every existing construction site (mechanical: drop trailing `false` arg, the top-level default does the rest).
- Streaming synchronizer holds a `latest_fdv1_fallback_` bool on `State`. Each `OnResponse` overwrites it from the incoming headers, so a fresh response without the directive clears it. Events between two responses share the flag from the most recent response.

## Test plan

- [x] New `fdv2_polling_impl_test.cpp` covers propagation across all variants + value comparison
- [x] New streaming tests cover wiring + clearing on subsequent response
- [x] Full server + SSE suites green

---

> [!NOTE]
> **Medium Risk**
> Changes the FDv2 source result contract and how sync responses are interpreted ahead of orchestrator fallback logic; mistakes could mis-signal FDv1 downgrade, but behavior is not yet wired to switching data modes.
> 
> **Overview**
> **FDv2 sources now surface a server-driven FDv1 fallback signal on every `FDv2SourceResult`.** The `fdv1_fallback` flag moves from nested variant payloads to a single top-level field (default `false`), reflecting that it comes from the HTTP response, not the result kind.
> 
> **Polling** reads `X-LD-FD-Fallback` (value `true`, case-insensitive) from poll HTTP headers and sets `fdv1_fallback` on 304/200 outcomes and HTTP error paths; transport-level failures without a response still leave it false.
> 
> **Streaming** registers an SSE `on_response` hook, keeps `latest_fdv1_fallback_` from the most recent response headers (cleared when a later response omits the header), and applies it in `Notify` to all delivered results—including errors after a directive.
> 
> Tests were updated for the new shape; new coverage exercises polling and streaming propagation. **Orchestrator behavior is unchanged** in this PR—the flag is stamped only for a follow-up to consume.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8e354cdae412d4d7555a6f79319438225ee6ffa4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/libs/server-sdk/src/data_systems/fdv2/fdv2_polling_impl.cpp
+++ b/libs/server-sdk/src/data_systems/fdv2/fdv2_polling_impl.cpp
@@ -4,6 +4,7 @@
 #include <launchdarkly/network/http_error_messages.hpp>
 #include <launchdarkly/server_side/config/builders/all_builders.hpp>
 
+#include <boost/algorithm/string/predicate.hpp>
 #include <boost/json.hpp>
 #include <boost/url/parse.hpp>
 #include <boost/url/url.hpp>
@@ -11,6 +12,7 @@
 namespace launchdarkly::server_side::data_systems {
 
 static char const* const kFDv2PollPath = "/sdk/poll";
+static char const* const kFDv1FallbackHeader = "X-LD-FD-Fallback";
 
 static char const* const kErrorParsingBody =
     "Could not parse FDv2 polling response";
@@ -30,6 +32,12 @@ static ErrorInfo MakeError(ErrorKind kind,
                            std::string message) {
     return ErrorInfo{kind, status, std::move(message),
                      std::chrono::system_clock::now()};
+}
+
+static bool ReadFDv1FallbackDirective(
+    network::HttpResult::HeadersType const& headers) {
+    auto const it = headers.find(kFDv1FallbackHeader);
+    return it != headers.end() && boost::iequals(it->second, "true");
 }
 
 network::HttpRequest MakeFDv2PollRequest(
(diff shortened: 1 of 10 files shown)


libs/server-sdk/src/data_systems/fdv2/fdv2_polling_impl.cpp
```diff
@@ -4,13 +4,15 @@
...
 
+#include <boost/algorithm/string/predicate.hpp>
 #include <boost/json.hpp>
 #include <boost/url/parse.hpp>
 #include <boost/url/url.hpp>
 
 namespace launchdarkly::server_side::data_systems {
 
 static char const* const kFDv2PollPath = "/sdk/poll";
+static char const* const kFDv1FallbackHeader = "X-LD-FD-Fallback";
```
kinyoklion, line 15: nit: Maybe this can be shred with streaming someplace.

kinyoklion (approved):

libs/server-sdk/src/data_systems/fdv2/fdv2_polling_impl.cpp
beekld: I briefly had that, but there wasn't an obvious place to put it, and I didn't want to add a new header for it. I will consider it in a follow-up.

PR #543, opened 2026-06-02
feat: add internal BigSegmentStoreWrapper with caching and status polling

beekld:
## Summary

Internal layer between the evaluator and a customer-provided
`IBigSegmentStore`: hashes context keys, caches membership lookups, polls the
store's metadata in the background to track availability/staleness, and
broadcasts status changes. No caller yet — the evaluator and the public status
provider come later in the stack.

## Design notes

- Context keys are hashed to `base64(sha256(key))` using the standard base64
  alphabet (not the existing URL-safe `Base64UrlEncode`), so the lookup key
  matches what the Relay Proxy writes.
- Concurrent misses for the same key coalesce into a single store query rather
  than each hitting the store.
- The background poll is built on `async::Delay` + `CancellationSource`.
- A store error encountered during a membership lookup flips the store status to
  unavailable. The spec requires this; the Java and Go SDKs don't do it, so this
  is a deliberate divergence.
- Changes `StoreMetadata::last_up_to_date` from a bare `milliseconds` to a
  `system_clock::time_point`, so staleness math can't mix epochs. This touches
  the Redis and DynamoDB stores (one line each) and their metadata tests.

## Test plan

- [x] Added unit tests
- [x] Full server-sdk suite green (499/499)
- [x] DynamoDB source suite green (40/40) against DynamoDB Local
- [x] Redis source suite green (37/37) against a live Redis

---

> [!NOTE]
> **Medium Risk**
> Changes the public `IBigSegmentStore` contract (`noexcept`, `StoreMetadata` type) and adds threaded/async polling plus store-error handling that affects how staleness and availability are reported; evaluator is not wired yet, limiting immediate eval impact.
> 
> **Overview**
> Adds **`BigSegmentStoreWrapper`**, the internal layer in front of `IBigSegmentStore`: hashes context keys as **`base64(sha256(key))`** via new **`Base64Encode`** (RFC 4648, not URL-safe), uses an LRU/TTL **`MembershipCache`**, coalesces concurrent lookups for the same key, polls metadata on a timer for availability/staleness, and broadcasts status changes. Membership lookups return **`BigSegmentsStatus`** (healthy/stale/store error); a failed lookup marks the store unavailable (per spec).
> 
> **`StoreMetadata::last_up_to_date`** is now a **`system_clock::time_point`** instead of raw milliseconds; Redis and DynamoDB stores and their tests are updated accordingly. **`IBigSegmentStore::GetMembership` / `GetMetadata`** are **`noexcept`**, with DynamoDB wrapping AWS calls in try/catch.
> 
> Supporting pieces: **`big_segments_status.hpp`**, **`#include <variant>`** in `promise.hpp`, and unit tests for the wrapper, cache, and standard base64.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0b492dd143275828c1ca85c5fc5a1ff1fb6d8a0c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/libs/server-sdk-dynamodb-source/src/dynamodb_big_segment_store.cpp
+++ b/libs/server-sdk-dynamodb-source/src/dynamodb_big_segment_store.cpp
@@ -65,99 +65,114 @@ DynamoDBBigSegmentStore::~DynamoDBBigSegmentStore() = default;
 
 IBigSegmentStore::GetMembershipResult DynamoDBBigSegmentStore::GetMembership(
     std::string const& context_hash) const noexcept {
-    Aws::DynamoDB::Model::GetItemRequest request;
-    request.SetTableName(table_name_);
-    request.SetConsistentRead(true);
-    request.AddKey(kPartitionKey,
-                   Aws::DynamoDB::Model::AttributeValue{user_namespace_});
-    request.AddKey(kSortKey,
-                   Aws::DynamoDB::Model::AttributeValue{context_hash});
+    try {
+        Aws::DynamoDB::Model::GetItemRequest request;
+        request.SetTableName(table_name_);
+        request.SetConsistentRead(true);
+        request.AddKey(kPartitionKey,
+                       Aws::DynamoDB::Model::AttributeValue{user_namespace_});
+        request.AddKey(kSortKey,
+                       Aws::DynamoDB::Model::AttributeValue{context_hash});
 
-    auto outcome = client_->GetItem(request);
-    if (!outcome.IsSuccess()) {
-        return tl::make_unexpected(outcome.GetError().GetMessage());
+        auto outcome = client_->GetItem(request);
+        if (!outcome.IsSuccess()) {
+            return tl::make_unexpected(outcome.GetError().GetMessage());
+        }
+
+        auto const& item = outcome.GetResult().GetItem();
+
+        std::vector<std::string> included;
+        std::vector<std::string> excluded;
+
+        // GetSS() silently returns an empty vector if the attribute is not
+        // actually a String Set, so check the type explicitly before reading.
+        if (auto const it = item.find(kBigSegmentsIncludedAttribute);
+            it != item.end()) {
+            if (it->second.GetType() !=
+                Aws::DynamoDB::Model::ValueType::STRING_SET) {
+                return tl::make_unexpected(
+                    std::string("DynamoDB Big Segments '") +
+                    kBigSegmentsIncludedAttribute +
+                    "' is not of type STRING_SET");
+            }
+            for (auto const& ref : it->second.GetSS()) {
+                included.emplace_back(ref);
+            }
+        }
+        if (auto const it = item.find(kBigSegmentsExcludedAttribute);
+            it != item.end()) {
+            if (it->second.GetType() !=
+                Aws::DynamoDB::Model::ValueType::STRING_SET) {
+                return tl::make_unexpected(
+                    std::string("DynamoDB Big Segments '") +
+                    kBigSegmentsExcludedAttribute +
+                    "' is not of type STRING_SET");
+            }
+            for (auto const& ref : it->second.GetSS()) {
+                excluded.emplace_back(ref);
+            }
+        }
+
+        return Membership::FromSegmentRefs(included, excluded);
+    } catch (std::exception const& e) {
+        return tl::make_unexpected(e.what());
     }
-
-    auto const& item = outcome.GetResult().GetItem();
-
-    std::vector<std::string> included;
-    std::vector<std::string> excluded;
-
-    // GetSS() silently returns an empty vector if the attribute is not
-    // actually a String Set, so check the type explicitly before reading.
-    if (auto const it = item.find(kBigSegmentsIncludedAttribute);
-        it != item.end()) {
-        if (it->second.GetType() !=
-            Aws::DynamoDB::Model::ValueType::STRING_SET) {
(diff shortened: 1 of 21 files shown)


libs/server-sdk/src/data_components/big_segments/big_segment_store_wrapper.cpp
```diff
@@ -0,0 +1,194 @@
...
+
+    if (!is_leader) {
+        std::unique_lock lock(query->mutex);
+        query->cv.wait(lock, [&query] { return query->result.has_value(); });
+        return *query->result;
+    }
+
+    // Query the store outside any lock, then publish the result to waiters and
+    // drop the in-flight entry.
+    auto result = store_->GetMembership(HashContextKey(context_key));
```
kinyoklion, line 117: Does IBigSegmentStore::GetMembership need to be noexcept?

libs/server-sdk/src/data_components/big_segments/big_segment_store_wrapper.cpp
```diff
@@ -0,0 +1,194 @@
...
+                    self->ScheduleNextPoll();
+                }
+                return {};
+            },
+            async::kInlineExecutor);
+}
+
+BigSegmentStoreStatus BigSegmentStoreWrapper::GetStatus() {
+    {
+        std::lock_guard lock(status_mutex_);
```
kinyoklion, line 164: Probably not a major concern, but should this have the same type of in-flight check as membership?

(It appears implementation side this is inconsistent, go/java just let it thunder while .Net creates a task and then re-uses it for subsequent waiters.)

libs/server-sdk/src/data_components/big_segments/big_segment_store_wrapper.hpp
```diff
@@ -0,0 +1,154 @@
...
+    /**
+     * @brief Returns the current store health. If no metadata poll has
+     * completed yet, performs one synchronously on the calling thread first, so
+     * the first caller still gets an accurate staleness reading.
+     */
+    [[nodiscard]] BigSegmentStoreStatus GetStatus();
+
+    /**
+     * @brief Registers a listener invoked whenever the store status changes
+     * (not on every poll). The returned connection unregisters on destruction.
```
kinyoklion: Does the connection actually un-register on destruction?

libs/server-sdk-dynamodb-source/src/dynamodb_big_segment_store.cpp
```diff
@@ -102,10 +101,9 @@ IBigSegmentStore::GetMembershipResult DynamoDBBigSegmentStore::GetMembership(
         it != item.end()) {
         if (it->second.GetType() !=
             Aws::DynamoDB::Model::ValueType::STRING_SET) {
-            return tl::make_unexpected(
-                std::string("DynamoDB Big Segments '") +
-                kBigSegmentsExcludedAttribute +
-                "' is not of type STRING_SET");
```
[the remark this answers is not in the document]
beekld: Done. I don't think it ever would've thrown, but better to err on the safe side, I suppose.

libs/server-sdk/src/data_components/big_segments/big_segment_store_wrapper.cpp
beekld: Seems reasonable. Done.

libs/server-sdk/src/data_components/big_segments/big_segment_store_wrapper.cpp
```diff
@@ -0,0 +1,198 @@
...
+    {
+        std::lock_guard lock(load_mutex_);
+        in_flight_.erase(context_key);
+    }
+    {
+        std::lock_guard lock(query->mutex);
+        query->result = result;
+    }
+    query->cv.notify_all();
+    return result;
```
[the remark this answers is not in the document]
beekld: I don't really think it can throw in anything other than, like, out of memory situations, but it's fair to err on the safe side. Fixed.

(This makes me miss `absl::Cleanup`.

libs/server-sdk/src/data_components/big_segments/big_segment_store_wrapper.cpp
beekld: Done, since it was easy.

libs/server-sdk/src/data_components/big_segments/big_segment_store_wrapper.hpp
beekld: The comment was incorrect. Fixed.

libs/server-sdk/src/data_components/big_segments/big_segment_store_wrapper.cpp
```diff
@@ -0,0 +1,198 @@
...
+                << "Big Segment store returned error: " << result.error();
+            MarkStoreUnavailable();
+            return {integrations::Membership::FromSegmentRefs({}, {}),
+                    BigSegmentsStatus::kStoreError};
+        }
+        membership = *result;
+    }
+
+    auto const status = GetStatus().stale ? BigSegmentsStatus::kStale
+                                          : BigSegmentsStatus::kHealthy;
```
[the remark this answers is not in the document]
beekld: I think the existing code is correct, looking at the spec and the java implementation.

kinyoklion (approved):

PR #547, opened 2026-06-04
fix: FDv2 fixes surfaced by v3 contract test harness

beekld:
FDv2 fixes for issues found running the v3 contract test harness against the server SDK.

- Polling URL trailing-slash handling.
- Filter key validation in FDv2 polling and streaming.
- Protocol handler accepts incremental updates after payload-transferred.
- Stream restart on parse failures and protocol errors.
- Unknown SSE event types skipped without parsing.
- Error events preserve protocol state (only accumulated changes cleared).

---

> [!NOTE]
> **Medium Risk**
> Changes core FDv2 sync behavior (multi-payload streams, error recovery, and reconnect triggers); mistakes could cause missed updates or extra reconnects, but scope is FDv2 data paths with expanded unit tests.
> 
> **Overview**
> **FDv2 protocol handler** now matches streaming incremental semantics: after `payload-transferred` it stays active in `kPartial` instead of resetting, so later put/delete + `payload-transferred` cycles emit **partial** changesets without a new `server-intent`. Put/delete are no longer dropped solely because state was inactive; **`error`** clears in-flight changes but keeps protocol state so updates can continue. Adds **`IsKnownEvent`** for spec-defined ignore behavior.
> 
> **Polling and streaming** build `/sdk/poll` and `/sdk/stream` by stripping a trailing-slash empty URL segment (avoids `//`), and apply the same **`ValidateFilterKey`** used elsewhere—invalid filters are omitted with an error log and a full-environment request. **`MakeFDv2PollRequest`** takes a **`Logger`** for that path.
> 
> **FDv2 streaming** skips unknown SSE event types before JSON parsing, and calls **`async_restart`** on malformed JSON or non-server **`HandleEvent`** protocol errors so the next connection starts clean.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3488e09ff9888b81eee199a2bb4818dbdee07392. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/libs/server-sdk/src/data_systems/fdv2/fdv2_polling_impl.cpp
+++ b/libs/server-sdk/src/data_systems/fdv2/fdv2_polling_impl.cpp
@@ -1,4 +1,5 @@
 #include "fdv2_polling_impl.hpp"
+#include "../background_sync/detail/payload_filter_validation/payload_filter_validation.hpp"
 #include "fdv2_changeset_translation.hpp"
 
 #include <launchdarkly/network/http_error_messages.hpp>
@@ -11,7 +12,6 @@
 
 namespace launchdarkly::server_side::data_systems {
 
-static char const* const kFDv2PollPath = "/sdk/poll";
 static char const* const kFDv1FallbackHeader = "X-LD-FD-Fallback";
 
 static char const* const kErrorParsingBody =
@@ -54,11 +54,14 @@ network::HttpRequest MakeFDv2PollRequest(
     }
 
     boost::urls::url u = parsed.value();
-    u.set_path(u.path() + kFDv2PollPath);
+    // Use segments to join the path; string concatenation would produce a
+    // double slash when the base URL ends in '/'.
+    u.segments().push_back("sdk");
+    u.segments().push_back("poll");
     if (selector.value) {
         u.params().append({"basis", selector.value->state});
     }
-    if (filter_key) {
+    if (filter_key && detail::ValidateFilterKey(*filter_key)) {
         u.params().append({"filter", *filter_key});
     }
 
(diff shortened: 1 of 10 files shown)


libs/server-sdk/src/data_systems/fdv2/fdv2_polling_impl.cpp
```diff
@@ -54,12 +55,25 @@ network::HttpRequest MakeFDv2PollRequest(
     }
 
     boost::urls::url u = parsed.value();
-    u.set_path(u.path() + kFDv2PollPath);
+    // A trailing '/' on the base URL appears as an empty final segment;
+    // remove it so subsequent push_backs don't produce a double slash.
+    auto segs = u.segments();
```
kinyoklion, line 60: nit: Maybe this trimming could be a utility function?

libs/server-sdk/src/data_systems/fdv2/fdv2_polling_impl.cpp
```diff
@@ -54,12 +55,25 @@ network::HttpRequest MakeFDv2PollRequest(
...
+        segs.pop_back();
+    }
+    segs.push_back("sdk");
+    segs.push_back("poll");
     if (selector.value) {
         u.params().append({"basis", selector.value->state});
     }
     if (filter_key) {
-        u.params().append({"filter", *filter_key});
+        if (detail::ValidateFilterKey(*filter_key)) {
```
kinyoklion, line 70: nit: Pre-existing, but this probably logs every poll?

kinyoklion (approved):

libs/server-sdk/src/data_systems/fdv2/fdv2_polling_impl.cpp
beekld: I considered it, but I just don't think it's worth it for the few places it's used.

libs/server-sdk/src/data_systems/fdv2/fdv2_polling_impl.cpp
beekld: Yes, and that was intentional, per the spec. I am happy to revisit it in the future.

PR #550, opened 2026-06-10
feat: evaluate big segments

beekld:
# Evaluate big segments

Wires the big segment store wrapper into the server-side evaluator so flags referencing big segments actually evaluate, and annotates the reason with the big segments status. Builds on SDK-2366, which introduced the `EvaluationReason::BigSegmentsStatus` type.

## What's in

- Big segment matching in `rules.cpp`:
  - looks up membership for the unbounded context kind
  - tests the `<key>.g<generation>` segment ref
  - short-circuits to non-match on a store error
  - falls through to the segment's rules on a membership miss
- `EvaluationStack` carries the per-evaluation status, membership cache, and per-key store-error set, so a context key is queried at most once per evaluation.
- The aggregated status is stamped onto the returned `EvaluationReason`.

## Design notes

When one evaluation queries multiple big segments with differing statuses, the least-trustworthy status wins (`NOT_CONFIGURED` > `STORE_ERROR` > `STALE` > `HEALTHY`). This follows the spec but diverges from Java/Go, which take the last status written.

## Testing

Added unit tests covering each evaluation branch against an in-memory store.

---

> [!NOTE]
> **Medium Risk**
> Changes core segment-matching and evaluation-reason behavior for flags using big segments; incorrect membership or status aggregation could change targeting outcomes, though behavior is heavily unit-tested.
> 
> **Overview**
> **Big segment evaluation** is wired into the server-side flag evaluator: unbounded segments now resolve membership via `BigSegmentStoreWrapper` instead of always returning non-match with a TODO.
> 
> `EvaluationStack` is extended with a non-owning store pointer, per-evaluation membership cache (one store lookup per context key), per-key store-error short-circuit, and aggregated **BigSegmentsStatus** using worst-wins precedence (`NOT_CONFIGURED` > `STORE_ERROR` > `STALE` > `HEALTHY`). `Evaluator` accepts an optional big-segment store, seeds the stack from it, and stamps the aggregated status onto the returned `EvaluationReason` when any big segment was consulted.
> 
> In `Contains` / `MatchBigSegment`, big segments use the unbounded context kind’s key, check `<segmentKey>.g<generation>`, treat store errors as non-match without retry, skip regular include/exclude lists, and fall through to segment rules on membership miss. New gtest coverage exercises not-configured, healthy/stale/error paths, precedence, caching, and rule fallthrough.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9b37b9a8de84937aacfd79117888a67cca88ef1f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/libs/server-sdk/src/evaluation/evaluator.cpp
+++ b/libs/server-sdk/src/evaluation/evaluator.cpp
@@ -53,8 +53,9 @@ EvaluationDetail<Value> Evaluator::Evaluate(
     EventScope const& event_scope) {
     EvaluationStack stack{big_segment_store_};
     auto detail = Evaluate(std::nullopt, flag, context, stack, event_scope);
-    if (auto status = stack.BigSegmentsStatus()) {
-        return WithBigSegmentsStatus(std::move(detail), *status);
+    auto status = stack.BigSegmentsStatus();
+    if (status != EvaluationReason::BigSegmentsStatus::kNone) {
+        return WithBigSegmentsStatus(std::move(detail), status);
     }
     return detail;
 }
--- a/libs/server-sdk/src/evaluation/evaluator.cpp
+++ b/libs/server-sdk/src/evaluation/evaluator.cpp
@@ -20,8 +20,26 @@ std::optional<std::size_t> TargetMatchVariation(
     launchdarkly::Context const& context,
     Flag::Target const& target);
 
-Evaluator::Evaluator(Logger& logger, data_interfaces::IStore const& source)
-    : logger_(logger), source_(source) {}
+namespace {
+EvaluationDetail<Value> WithBigSegmentsStatus(EvaluationDetail<Value> detail,
+                                              enum EvaluationReason::BigSegmentsStatus status) {
+    auto const& maybe_reason = detail.Reason();
+    if (!maybe_reason) {
+        return detail;
+    }
+    EvaluationReason const& reason = *maybe_reason;
+    EvaluationReason updated(
+        reason.Kind(), reason.ErrorKind(), reason.RuleIndex(), reason.RuleId(),
+        reason.PrerequisiteKey(), reason.InExperiment(), status);
+    return EvaluationDetail<Value>(detail.Value(), detail.VariationIndex(),
+                                   std::move(updated));
+}
+}  // namespace
+
+Evaluator::Evaluator(Logger& logger,
+                     data_interfaces::IStore const& source,
+                     data_components::BigSegmentStoreWrapper* big_segment_store)
+    : logger_(logger), source_(source), big_segment_store_(big_segment_store) {}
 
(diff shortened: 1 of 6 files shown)


libs/server-sdk/src/evaluation/evaluator.cpp
```diff
@@ -20,8 +20,26 @@ std::optional<std::size_t> TargetMatchVariation(
...
-    : logger_(logger), source_(source) {}
+namespace {
+EvaluationDetail<Value> WithBigSegmentsStatus(EvaluationDetail<Value> detail,
+                                              enum EvaluationReason::BigSegmentsStatus status) {
+    auto const& maybe_reason = detail.Reason();
+    if (!maybe_reason) {
+        return detail;
+    }
+    EvaluationReason const& reason = *maybe_reason;
+    EvaluationReason updated(
```
kinyoklion, line 31: nit: Part of me wants this to be more colocated with the evaluation. But I suppose a real constructor change would be noticed.

libs/server-sdk/src/evaluation/rules.cpp
```diff
@@ -161,8 +239,11 @@ tl::expected<bool, Error> Contains(Segment const& segment,
     }
 
     if (segment.unbounded) {
-        // TODO(sc209881): set big segment status to NOT_CONFIGURED.
-        return false;
+        if (auto match = MatchBigSegment(segment, context, stack)) {
+            return *match;
+        }
+        // The membership had no entry for this segment; fall through to its
```
kinyoklion: In the case where we have an unbounded segment we want to skip the include/exclude of a normal segment and go directly to rules.

> When a segment is loaded during a SEGMENT_MATCH operation and its unbounded property is true, the big segment store MUST be used in place of the segment's included, excluded, includedContexts, and excludedContexts fields for determining membership.

But I confirmed against a few SDKs as well.

https://github.com/launchdarkly/go-server-sdk-evaluation/blob/cc2c96e2119aa953304df3fbff1a8033c82e6a50/evaluator_segment.go#L32-L101

https://github.com/launchdarkly/java-core/blob/30cf64b9ab81643b511835005117532bdfd353a2/lib/sdk/server/src/main/java/com/launchdarkly/sdk/server/Evaluator.java#L473-L521

https://github.com/launchdarkly/dotnet-core/blob/e41f1666ff704dc4b4534ebba3d92a20a0d7516a/pkgs/sdk/server/src/Internal/Evaluation/EvaluatorSegment.cs#L27-L67

beekld: Yes, this was the bug I mentioned in chat. Fixed.

kinyoklion (approved):

PR #552, opened 2026-06-11
feat: expose Big Segments via public Client and Config APIs

beekld:
# Expose Big Segments to customer code

Makes the server-side Big Segments feature reachable from customer code: a public config method to wire a store, and a status provider on the client for querying store health and subscribing to changes. Builds on SDK-2366-2.

## What's in

- Promotes `BigSegmentsBuilder` to the public include tree and adds `ConfigBuilder::BigSegments(BigSegmentsBuilder)`, surfaced as an optional `Config::BigSegments()`.
- New public `IBigSegmentStoreStatusProvider` and `BigSegmentStoreStatus` (`IsAvailable()`/`IsStale()`), plus `Client::BigSegmentStoreStatus()`.
- `ClientImpl` constructs the `BigSegmentStoreWrapper` from config, hands it to the evaluator, and starts its background poll. An internal adapter exposes the wrapper's status broadcaster as the public provider.
- `BigSegmentsBuilder::Build()` now returns a new error code when the store is null.

## Design notes

- The status surface mirrors `DataSourceStatus`: the accessor is `Status()`, and listeners use `OnBigSegmentStoreStatusChange(handler) -> IConnection`.
- When no store is configured, the provider reports the store as unavailable and registered listeners never fire.

## Testing

Unit tests added for the new behavior.

---

> [!NOTE]
> **Medium Risk**
> Changes flag evaluation wiring and add a new public config surface; misconfiguration or store outages can affect Big Segment targeting, though behavior is gated on optional config and mirrors existing data-source status patterns.
> 
> **Overview**
> This PR wires **Big Segments** into the public server SDK: customers can attach a store via config and observe store health on the client, similar to data source status.
> 
> **Configuration:** `ConfigBuilder::BigSegments(BigSegmentsBuilder)` stores an optional `BigSegmentsConfig` on `Config`. `BigSegmentsBuilder::Build()` now returns `tl::expected` and fails with `kConfig_BigSegments_NullStore` when the store pointer is null (previously a null store was allowed through).
> 
> **Client runtime:** When Big Segments are configured, `ClientImpl` creates `BigSegmentStoreWrapper`, passes it to the evaluator, starts background metadata polling, and exposes status through `Client::BigSegmentStoreStatus()`.
> 
> **Public status API:** New `BigSegmentStoreStatus` (`IsAvailable` / `IsStale`) and `IBigSegmentStoreStatusProvider` with `Status()` and `OnBigSegmentStoreStatusChange` → `IConnection`. An internal adapter maps the wrapper to this API; with no store configured, status is unavailable/not stale and listeners are no-ops.
> 
> Unit tests cover builder validation, config round-trip, and status provider behavior (including transitions).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 11902fa852d36bf17345153bc0eabbb0469e053a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/libs/server-sdk/src/data_components/big_segments/big_segment_store_status_provider.cpp
+++ b/libs/server-sdk/src/data_components/big_segments/big_segment_store_status_provider.cpp
@@ -1,0 +1,50 @@
+#include "big_segment_store_status_provider.hpp"
+
+#include <utility>
+
+namespace launchdarkly::server_side::data_components {
+
+namespace {
+
+// Returned when no store is configured: nothing to disconnect.
+class NoopConnection : public IConnection {
+   public:
+    void Disconnect() override {}
+};
+
+// Converts the internal status struct to the public status type. Both are named
+// BigSegmentStoreStatus, in this namespace and in server_side respectively.
+server_side::BigSegmentStoreStatus ToPublic(
+    data_components::BigSegmentStoreStatus const& status) {
+    return server_side::BigSegmentStoreStatus{status.available, status.stale};
+}
+
+}  // namespace
+
+BigSegmentStoreStatusProvider::BigSegmentStoreStatusProvider(
+    std::shared_ptr<BigSegmentStoreWrapper> wrapper)
+    : wrapper_(std::move(wrapper)) {}
+
+server_side::BigSegmentStoreStatus BigSegmentStoreStatusProvider::Status()
+    const {
+    if (!wrapper_) {
+        return server_side::BigSegmentStoreStatus{/* available= */ false,
+                                                  /* stale= */ false};
+    }
+    return ToPublic(wrapper_->GetStatus());
+}
+
+std::unique_ptr<IConnection>
+BigSegmentStoreStatusProvider::OnBigSegmentStoreStatusChange(
+    std::function<void(server_side::BigSegmentStoreStatus status)> handler) {
+    if (!wrapper_) {
+        return std::make_unique<NoopConnection>();
+    }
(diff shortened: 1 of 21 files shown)


libs/server-sdk/src/data_components/big_segments/big_segment_store_status_provider.cpp
```diff
@@ -0,0 +1,50 @@
...
+std::unique_ptr<IConnection>
+BigSegmentStoreStatusProvider::OnBigSegmentStoreStatusChange(
+    std::function<void(server_side::BigSegmentStoreStatus status)> handler) {
+    if (!wrapper_) {
+        return std::make_unique<NoopConnection>();
+    }
+    return wrapper_->OnStatusChange(
+        [handler = std::move(handler)](
+            data_components::BigSegmentStoreStatus status) {
+            handler(ToPublic(status));
```
kinyoklion, line 46: Do we need some exception wrapping around the handler?

kinyoklion (approved):

libs/server-sdk/src/data_components/big_segments/big_segment_store_status_provider.cpp
beekld: None of the other public APIs that take callbacks seem to wrap the handlers, so I don't think so. We could do a pass and add it to all of them if we think it's a good idea.

PR #582, opened 2026-07-21
fix: add defensive cycle guard to prerequisite evaluation

tanderson-ld:
## Summary

Adds defensive cycle detection to the recursive prerequisite walk in `ClientImpl::VariationInternal`, bringing the C++ client SDK's behavior into line with the LaunchDarkly server SDK evaluators.

Tracker: [SDK-2709](https://launchdarkly.atlassian.net/browse/SDK-2709). Parent: [SDK-2695](https://launchdarkly.atlassian.net/browse/SDK-2695). Spec change: [sdk-specs#246](https://github.com/launchdarkly/sdk-specs/pull/246) (CSPE 1.2.5, 1.2.5.1, 1.2.5.2). Contract-test coverage: [sdk-test-harness#384](https://github.com/launchdarkly/sdk-test-harness/pull/384) (merged in v2.38.0). Companion PRs: [android-client-sdk#377](https://github.com/launchdarkly/android-client-sdk/pull/377), [ios-client-sdk#512](https://github.com/launchdarkly/ios-client-sdk/pull/512), [js-core#1816](https://github.com/launchdarkly/js-core/pull/1816), [flutter-client-sdk#329](https://github.com/launchdarkly/flutter-client-sdk/pull/329), [dotnet-core#317](https://github.com/launchdarkly/dotnet-core/pull/317).

## Background

The LaunchDarkly service validates prerequisite graphs on every mutation and rejects any change that would produce a cycle, so under normal operation the SDK does not see a cyclic prerequisite graph. Server-side SDK evaluators nonetheless carry defensive cycle detection for exceptional cases — for example, delivery of updates out of order or a persisted state loaded from disk that predates a subsequent correction. This PR extends the same defensive posture to the C++ client SDK.

## The fix

`VariationInternal<T>` now threads a lazily-allocated `std::unordered_set<std::string>*` through the recursion carrying the flag keys on the current evaluation path. Before descending into a prerequisite, the walker checks whether that key is already on the path; if so, it skips that edge and continues with remaining prerequisites at the same level.

- **Lazy allocation**: variation calls on prereq-less flags (the common case) allocate zero collections. The `std::unordered_set` is created only when the walker descends into a prerequisite for the first time in a call, then reused for the rest of the walk.
- **Mutable `insert` / catch-and-rethrow `erase`**: no per-descent copy. `ancestors->insert(key)` before recursing; if a nested call throws, the exception handler restores `ancestors` to its pre-descent state before rethrowing. This preserves the invariant "the set contains exactly the current path" even under exceptions.
- **Ancestor-set (current-path) semantics** — this is deliberate. A prerequisite reached via multiple non-cyclic paths (a diamond `A -> [B, C], B -> [D], C -> [D]`) is not a cycle; each path should emit its own prerequisite event. A global visited-set would silently drop the second event; the ancestor-set pattern correctly emits D twice. The contract-test suite has a case guarding this.

The optional trailing parameter `std::unordered_set<std::string>* visited = nullptr` is added to `VariationInternal<T>`, so all existing callers (the public `BoolVariation` / `IntVariation` / `StringVariation` / `DoubleVariation` / `JsonVariation` and their `*Detail` counterparts) work unchanged. The recursive call now goes directly to `VariationInternal<Value>` rather than through `JsonVariation`, so the visited pointer can be threaded through — behavior is otherwise identical.

## Caller-visible behavior on a cycle

The requested flag returns its cached value and reason unchanged. A client-side prerequisite cycle is not surfaced as `MALFORMED_FLAG` and does not fall back to the caller-provided default value. This differs from server-side behavior — the client already holds an authoritative pre-evaluated result from the server; only the ancillary event walk is affected by the cycle.

## Files changed

- `libs/client-sdk/src/client_impl.hpp` — added `std::unordered_set<std::string>*` optional parameter to `VariationInternal<T>` declaration, plus the `<string>` / `<unordered_set>` includes.
- `libs/client-sdk/src/client_impl.cpp` — cycle guard implementation in `VariationInternal<T>`.
- `contract-tests/client-contract-tests/src/main.cpp` — declares `client-prereq-cycle-detection` capability so the matching contract tests in sdk-test-harness v2.38.0+ activate for this SDK.

## Verification

- **Build**: `cmake --build build --target launchdarkly-cpp-client` → clean, no warnings.
- **Contract tests locally** against harness v2.38.1: `client-tests` binary + `sdk-test-harness v2.38.1` → **876 total, 12 skipped, 864 ran, all passed.** Both new suites (`events/prerequisite events handle cycles` and `events/summary events/prerequisites/handles cycles`) fire all 15 subtests including the deep-chain non-cyclic control, no failures.

The C++ client SDK does not have existing unit tests that exercise prerequisite walks (`libs/client-sdk/tests/client_test.cpp` currently covers only default-value paths with no flag data loaded). Rather than build new unit-test infrastructure for this PR, verification is provided by the contract-test suite, which now exercises the fix comprehensively.

## Changelog

> Updated prerequisite evaluation event emission to match server SDK behavior for cyclic prerequisite graphs.

[SDK-2709]: https://launchdarkly.atlassian.net/browse/SDK-2709?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[SDK-2695]: https://launchdarkly.atlassian.net/browse/SDK-2695?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---

> [!NOTE]
> **Medium Risk**
> Touches core flag evaluation and analytics event emission for prerequisites; behavior change is scoped to exceptional cyclic graphs and is covered by contract tests.
> 
> **Overview**
> Adds **defensive cycle detection** to the recursive prerequisite walk in `ClientImpl::VariationInternal`, aligning client-side prerequisite **event emission** with server SDK behavior when a cyclic graph is encountered (e.g. stale persisted state).
> 
> `VariationInternal` now accepts an optional `std::unordered_set<std::string>*` tracking **ancestors on the current path** (not a global visited set, so diamond graphs still emit prerequisite events on each path). Before recursing into a prerequisite, if that key is already on the path the edge is **skipped** with a warning log; the requested flag’s cached value and reason are unchanged. Prerequisite recursion calls `VariationInternal<Value>` directly so the ancestor set can be threaded through.
> 
> Contract tests advertise the new **`client-prereq-cycle-detection`** capability in `client-contract-tests`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b9e27d5d151075339e6f31b50d038d2e5d9ffaed. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


contract-tests/server-contract-tests/test-suppressions.txt
tanderson-ld, line 1: Unrelated to this PR.  Needed to suppress these while @keelerm84 works through addressing these in the various affected SDKs.

contract-tests/client-contract-tests/src/main.cpp
tanderson-ld, line 1: Existing lint issues are being ignored by this PR.

contract-tests/server-contract-tests/test-suppressions.txt
keelerm84: Will be fixed by https://github.com/launchdarkly/cpp-sdks/pull/583

libs/client-sdk/src/client_impl.cpp
```diff
@@ -308,19 +309,34 @@ EvaluationDetail<T> ClientImpl::VariationInternal(FlagKey const& key,
...
+        std::unordered_set<std::string> local_visited;
+        auto* ancestors = visited ? visited : &local_visited;
+        ancestors->insert(key);
+        try {
+            for (auto const& prereq : *prereqs) {
+                if (ancestors->count(prereq) != 0) {
+                    // Cyclic edge: skip descent, continue with remaining
+                    // prerequisites at this level. The requested flag's value
+                    // and reason (below) are unaffected.
+                    continue;
```
beekld: Maybe we should log a warning here?

libs/client-sdk/src/client_impl.cpp
```diff
@@ -308,19 +309,34 @@ EvaluationDetail<T> ClientImpl::VariationInternal(FlagKey const& key,
     // prerequisites (recursively), which is necessary to ensure LaunchDarkly
     // analytics functions properly.
     //
-    // We're using JsonVariation because the type of the
-    // prerequisite is both unknown and irrelevant to emitting the events.
-    //
-    // We're passing Value::Null() to match a server-side SDK's behavior when
```
beekld, line 314: The part of this comment about Value::Null() is still valid, right? Maybe we should keep that part?

libs/client-sdk/src/client_impl.cpp
```diff
@@ -308,19 +309,34 @@ EvaluationDetail<T> ClientImpl::VariationInternal(FlagKey const& key,
...
-    if (auto const prereqs = flag.Prerequisites()) {
-        for (auto const& prereq : *prereqs) {
-            JsonVariation(prereq, Value::Null());
+    // refactoring the code below.
+    auto const prereqs = flag.Prerequisites();
+    if (prereqs && !prereqs->empty()) {
+        std::unordered_set<std::string> local_visited;
+        auto* ancestors = visited ? visited : &local_visited;
+        ancestors->insert(key);
+        try {
```
beekld: We don't really use exceptions in this repo except to `catch` them when calling out to third-party or user code that might throw weird ones. Anything our SDK itself might `throw` (like OOM) seems like the kind of fatal thing we wouldn't handle anyway. I would suggest just not wrapping it in a `try` at all.

beekld (approved):

libs/client-sdk/src/client_impl.cpp
tanderson-ld: Good call — added a warn-level log in [`b9e27d5d`](https://github.com/launchdarkly/cpp-sdks/pull/582/commits/b9e27d5d). Cycles shouldn't reach the SDK under normal operation (LD validates the graph on save), so if one does fire at runtime it's worth surfacing.

libs/client-sdk/src/client_impl.cpp
tanderson-ld: Good catch — I over-trimmed. Restored the `Value::Null()` sentence in [`b9e27d5d`](https://github.com/launchdarkly/cpp-sdks/pull/582/commits/b9e27d5d). Dropped only the sibling "we're using JsonVariation" sentence since we now call `VariationInternal<Value>` directly, so that part is stale.

libs/client-sdk/src/client_impl.cpp
tanderson-ld: Agreed — dropped the try/catch in [`b9e27d5d`](https://github.com/launchdarkly/cpp-sdks/pull/582/commits/b9e27d5d). `VariationInternal` is our own code and doesn't throw, and `local_visited` is a stack local that stack-unwind destroys on the exceptional path, so the erase-in-catch was unnecessary.

PR #594, opened 2026-08-12
feat: Add environment ID support for hooks.

kinyoklion:
## Summary

Server SDK: the `X-LD-EnvID` response header now reaches `EvaluationSeriesContext` / `TrackSeriesContext`, for both FDv1 and FDv2 data systems.

- The environment ID rides the data payload into the store: `SDKDataSet::environment_id` (FDv1 `Init`) and `ChangeSet::environment_id` (FDv2 `Apply`).
- `MemoryStore` retains it while applying the payload — absent/empty values leave a previously reported ID in place — and reports nothing before initialization.
- `IDataSystem::EnvironmentId()` is what `ClientImpl` passes into the hook contexts (`std::nullopt` for offline/lazy load).
- Only successful responses count; errors and missing headers leave the value untouched.

Screenshots and a staging preview don't apply: SDK-internal data path, no UI.

<details>
<summary>Implementation details</summary>

FDv1 polling attaches the header of a successful response to the parsed data set; FDv1 streaming attaches the connection's header to the `put` payload it hands to the destination; the FDv1 fallback adapter stamps it onto the change sets it synthesizes for FDv2, retaining it for subsequent partials.

```cpp
// FDv1 polling
poll_result->environment_id = ReadEnvironmentId(res.Headers());
sink_->Init(std::move(*poll_result));

// FDv2 polling / streaming
cs->change_set.environment_id = ReadEnvironmentId(res.Headers());
```

`kEnvironmentIdHeader` and the `ReadEnvironmentId` overloads (HTTP result and Beast headers) live in `data_systems/environment_id_header.hpp` and are shared by all four data sources.

Verified with harness `hooks/evaluation/provides the environment ID` (default + polling) against both v2.39.0 (FDv1) and v3.2.0-alpha.6 (FDv2), plus the server unit suite and the full v3 contract suite.

</details>

Link to Devin session: https://app.devin.ai/sessions/bfe54128e2804a96bb100e6120e9a3ef
Requested by: @kinyoklion


libs/server-sdk/src/data_systems/background_sync/sources/polling/polling_data_source.cpp
```diff
@@ -102,6 +102,13 @@ void PollingDataSource::DoPoll() {
 }
 
 void PollingDataSource::HandlePollResult(network::HttpResult const& res) {
+    if (!res.IsError() && (res.Status() == 200 || res.Status() == 304)) {
+        if (auto const it = res.Headers().find("X-LD-EnvID");
```
kinyoklion: We should have a centralized location for this header definition. Also potentially a shared function for getting this header and setting it into a sink.

libs/server-sdk/src/data_systems/background_sync/sources/streaming/streaming_data_source.cpp
```diff
@@ -125,16 +126,31 @@ void StreamingDataSource::StartAsync(
...
+            if (status ==
+                DataSourceEventHandler::MessageStatus::kInvalidMessage) {
                 // Invalid data received - restart the connection with backoff
                 // to get a fresh stream. The backoff mechanism prevents rapid
                 // reconnection attempts.
                 LD_LOG(self->logger_, LogLevel::kWarn)
-                    << "Received invalid data from stream, restarting connection";
+                    << "Received invalid data from stream, restarting "
+                       "connection";
                 if (self->client_) {
```
kinyoklion: Why were these lines changed?

kinyoklion:
We should refactor this to follow the patterns used by other SDKs. Where the environment ID is in the store and the store supports an interface that allows setting it with the payload.

